PR #204, opened 2018-04-06
Changed check in TerminateYarn to stop infinite loop

bennybechdk:
I have experienced infinete loops, because the Thread was stopped and terminated, but not free'ed or suspended.

### DIFF
--- a/Lib/Core/IdSchedulerOfThread.pas
+++ b/Lib/Core/IdSchedulerOfThread.pas
@@ -218,7 +218,7 @@ var
 begin
   Assert(AYarn<>nil);
   LYarn := TIdYarnOfThread(AYarn);
-  if (LYarn.Thread <> nil) and (not LYarn.Thread.Suspended) then begin
+  if (LYarn.Thread <> nil) and (not LYarn.Thread.Stopped) then begin
     // Is still running and will free itself
     LYarn.Thread.Stop;
     // Dont free the yarn. The thread frees it (IdThread.pas)


rlebeau:
I have not experienced infinite loops with the original code. Please provide an example scenario that causes an infinite loop. What environment does the scenario run in?  Calling TIdThread.Stop() does not resume a suspended thread, so I don't think the requested change is a proper fix for an infinite loop.  Please explain your reasoning.

bennybechdk:
Thats a fair, my explanation is quite thin.

I am using a Indy  HTTP server to receive a OAUTH2 SigninCode, using the standard browser for login, where the SigninCode is served as a param on a redirect to the indyhttpserver.
After the code is received, the program writes a response (and sets the connection to auto disconnect), and waits a until the data is flushed (can't remember the exact code, but i can find it).
Then the program set active on the server til false (and it is here the infinite loop is in TerminateAllYarns because the Thread is Stopped and not suspended), and frees the server.
When I looked at the code, the most logical solution (for me) was to check if the thread wasn't stopped (also includes suspended), then stop it.

For a very long time only a few of our Customers have experienced the problem (and I haven't experienced it),
Then Friday morning i saw it every time i did login. So i just had to find the problem and a solution.

bennybechdk:
**Example Code from the (sometimes) failing program:**

`
procedure TIPDOAuth2Authenticator.GetAuthorizationTokenFromIDP;
var
  webserver : TIdHTTPServer;
  VentIndtil : TDateTime;
  ContextCount : Integer;
begin
    webserver := TIdHTTPServer.Create(self);
    webserver.OnCommandGet := wsOnGetCommand;
    webserver.DefaultPort := 3100 + Random(200);
    repeat
      try
        webserver.Bindings.Clear;  // Otherwise, DefaultPort will not be used
        webserver.DefaultPort := webserver.DefaultPort + 1;
        webserver.Active := True;
      except
        on e: EIdCouldNotBindSocket  do  { Ignore error }      ;
        on e: EIdSocketError do
        if e.LastError <> 10048 { socket already in UseDockManager } then raise;
      end;
    until webserver.Active or (webserver.DefaultPort >= 3999);
    AuthOkRedirectURI := 'http://localhost:' + IntToStr(webserver.DefaultPort) + '/TOauth2LoginOk';
  OAuthWebForm := Tfrm_OAuthWebForm.Create(self);
  try
    OAuthWebForm.ShowWithURL(AuthorizationRequestURI, 'Login to T:  ');  // opens std browser on AuthorizationRequestURI
    if OAuthWebForm.ModalResult <> mrOK then
    begin
      // If you do not want to log in, the program must be closed.
      Application.Terminate;
      // Wait until the terminated flag is set correctly.
      while not Application.Terminated do Application.ProcessMessages;
    end;
      // wait between 0.3 and 2 seconds for connections to complete (and data has been written).
      VentIndtil := IncSecond(now, 2);
      Sleep(300);
      Application.ProcessMessages;
      repeat
        Sleep(20);
        Application.ProcessMessages;
        try
          ContextCount := webserver.Contexts.LockList.Count;
        finally
          webserver.Contexts.UnlockList;
        end;
      until (VentIndtil < now) or (ContextCount = 0);
      webserver.Active := false;
      webserver.Free;
    end;
  finally
    OAuthWebForm.Release;
    OAuthWebForm := nil;
  end;
end;

procedure TIPDOAuth2Authenticator.wsOnGetCommand(AContext: TIdContext;
    ARequestInfo: TIdHTTPRequestInfo; AResponseInfo: TIdHTTPResponseInfo);
var
  AutoCloseScript: string;
begin
    AResponseInfo.ContentText := '<html><head><title>Trio login OK</title></head>' +
      '<body>You are now logged in</body></html>';
    // Luk connection, when answer is send.
    AResponseInfo.CloseConnection := AResponseInfo.CloseConnection or (ARequestInfo.CommandType = hcGET);
    OAuthWebForm.BringToFront;
    OAuthWebForm.ModalResult := mrOk;
    Application.ProcessMessages; // update screen, to remove the closed OAuthWebForm.
 end;

`

rlebeau:
You are misusing `TIdHTTPServer`.

The `OnCommand` event handler is fired in a worker thread, not the main UI thread.  The event handler code you have shown is not thread-safe.  You cannot directly manipulate the UI in a worker thread like you are, that will cause all kinds of problems, including deadlocks, crashes, dead UIs, etc.  You MUST synchronize with the main UI thread when you want to update the UI (and get rid of the call to `Application.ProcessMessages()`, it doesn't belong).  This is the root of your problem.  I suggest using `TThread.Queue()` or `TIdNotify` or other asynchronous synchronization of your choosing, your `OnCommand` handler should not be waiting on the UI at all.

When you set the `TIdHTTPServer.Active` property to false, the property setter will block the calling thread until all worker threads have terminated, but your `OnCommand` code does not allow a worker thread to terminate correctly.  That is why your server is not shutting down.  It is not a bug in Indy itself.

Try something more like this instead:

```
procedure TIPDOAuth2Authenticator.GetAuthorizationTokenFromIDP;
var
  webserver : TIdHTTPServer;
  VentIndtil : TDateTime;
  ContextCount : Integer;
begin
  webserver := TIdHTTPServer.Create(nil);
  try
    webserver.OnCommandGet := wsOnGetCommand;
    webserver.DefaultPort := 3100 + Random(200);
    repeat
      try
        webserver.Bindings.Clear; // Otherwise, DefaultPort will not be used
        webserver.DefaultPort := webserver.DefaultPort + 1;
        webserver.Active := True;
      except
        on e: EIdCouldNotBindSocket do { Ignore error } ;
        on e: EIdSocketError do
          if e.LastError <> 10048 { socket already in Use } then raise;
      end;
    until webserver.Active or (webserver.DefaultPort > 3999);
    if not webserver.Active then
    begin
      // If you cannot log in, the program must be closed.
      Application.Terminate;
      Exit;
    end;
    AuthOkRedirectURI := 'http://localhost:' + IntToStr(webserver.DefaultPort) + '/TOauth2LoginOk';
    OAuthWebForm := Tfrm_OAuthWebForm.Create(nil);
    try
      OAuthWebForm.ShowWithURL(AuthorizationRequestURI, 'Login to T: '); // opens std browser on AuthorizationRequestURI
      if OAuthWebForm.ModalResult <> mrOK then
      begin
        // If you do not want to log in, the program must be closed.
        Application.Terminate;
      end;
    finally
      OAuthWebForm.Release;
      OAuthWebForm := nil;
    end;
  finally
    webserver.Free;
  end;
end;

procedure TIPDOAuth2Authenticator.wsOnGetCommand(AContext: TIdContext;
  ARequestInfo: TIdHTTPRequestInfo; AResponseInfo: TIdHTTPResponseInfo);
begin
  if ARequestInfo.Document = '/TOauth2LoginOk' then
  begin
    if ARequestInfo.CommandType = hcGET then
    begin
      AResponseInfo.ResponseNo := 200;
      AResponseInfo.ContentText := '<html><head><title>Trio login OK</title></head><body>You are now logged in</body></html>';
      AResponseInfo.ContentType := 'text/html';
      TThread.Queue(nil,
        procedure
        begin
          if OAuthWebForm <> nil then
          begin
            OAuthWebForm.BringToFront;
            OAuthWebForm.ModalResult := mrOk;
          end;
        end
      );
    end else begin
      AResponseInfo.ResponseNo := 405;
    end;
  end else begin
    AResponseInfo.ResponseNo := 404;
  end;
  AResponseInfo.CloseConnection := True;
end;
```

That being said, OAuth authentication does not require you to actually run an HTTP server at all, so you should just get rid of the `TIdHTTPServer` altogether.  You should embed a web browser directly in your UI and then detect when that browser attempts to redirect to your chosen URL.  For instance, VCL and FMX both have `TWebBrowser` components, with `OnBeforeNavigate2` and `OnShouldStartLoadWithRequest` events, respectively.  Or use a 3rd party browser component.  Either way, the URL you redirect OAuth to can be anything you want, it doesn't have to be a real live URL at all.  When you detect the embedded browser redirecting to your URL, simply stop it, close the browser, and update your UI as needed.

bennybechdk:
Thanks for Clarifying my misuse, I have now corrected it.
It was obvious (now).

About your suggestion with the embedded browser:
I started using an embedded browser, but some of my customers was using special browser plugins for the login and others were using single sign on (that didn’t work in embedded browser)
And if I am a little security focused, programs using embedded browsers can see (and steel) the users credentials). (and google has started blocking it https://auth0.com/blog/google-blocks-oauth-requests-from-embedded-browsers/).
That’s why I am using embedded webserver to get the response (option 2 in https://developers.google.com/identity/protocols/OAuth2InstalledApp).

Issue #87, opened 2017-04-06
TIdSSPINTLMAuthentication does not differentiate between Negotiate and NTLM authorizations


rlebeau:
TIdSSPINTLMAuthentication is registered for both "Negotiate" and "NTLM" HTTP authorizations, but when it formats an Authorization request, it always creates an "NTLM" request, never a "Negotiate" request. It needs to do so, even if Negotiate has NTLM enabled.

tothpaul:
I have a 407 error until I open a random website under IE, then everything works fine until probably a kind of timeout ... it this issue about that ?
Thanks

PR #235, opened 2018-10-31
Fix for TIdSSPINTLMAuthentication

markjforte:
TIdSSPINTLMAuthentication currently does not work.  This appears to have been an issue for quite some time as I found other folks that appear to have had the same issue way back in 2012:
http://codeverge.com/embarcadero.delphi.winsock/tidhttp-not-retrying-after-407-res/1073579
Here is another thread that seems to be related to the issue:
https://stackoverflow.com/questions/15111605/indy-tidhttp-integrated-security-authentication-issue-with-sspi
Note that there is an existing open issue, but I'm not sure if my fix addresses it, or if it is for something different?:
https://github.com/IndySockets/Indy/issues/87
Here is what I changed:
IdAuthentication.pas - Added virtual method TIdAuthentication.AuthorizationAttemptMerited.  Contains exact same logic that used to be TIdCustomHTTP.DoOnProxyAuthorization.
TIdHTTP.pas - Updated TIdCustomHTTP.DoOnProxyAuthorization to call TIdAuthentication.AuthorizationAttemptMerited rather than doing the logic itself.
IdAuthenticationSSPI.pas - TIdSSPINTLMAuthentication.AuthorizationAttemptMerited overrides default logic.  For this class authorization is always merited as no username or password is needed.

### DIFF
--- a/Lib/Protocols/IdAuthentication.pas
+++ b/Lib/Protocols/IdAuthentication.pas
@@ -85,6 +85,8 @@ type
 
     function Authentication: String; virtual; abstract;
     function KeepAlive: Boolean; virtual;
+    function AuthorizationAttemptMerited(
+      const AProxyPasswordeExists, AOnProxyAuthorizationExists: Boolean): Boolean; virtual;
     function Next: TIdAuthWhatsNext;
 
     property AuthParams: TIdHeaderList read FAuthParams write SetAuthParams;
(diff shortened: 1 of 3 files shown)


markjforte:
I want to update my commit comments.  I will submit new pull request shortly.

rlebeau:
Without altering TIdHTTP's source code, another way to approach this is to simply assign an OnProxyAuthorization handler to TIdHTTP and have it return Handled=True even though no username/password is blank.  TIdSSPINTLMAuthenticate.DoNext() already handles that case.

markjforte:
Hi Remy, after several hours of diagnosing and figuring out what was going on I did realize that I could do that workaround.  But I figured that others might benefit if this could be fixed in the library (save them from going through the frustration that I went through), so I decided to put this out there.  Thanks!

rlebeau:
I'm not really liking the proposed approahc of adding a virtual method to TIdAuthentication.

But, it might be worth to instead tweak TIdCustomHTTP.DoOnProxyAuthorization() to simply remove the initial check for OnProxyAuthorization and ProxyParams.ProxyPassword being assigned, and let TIdHTTP call LAuth.Next unconditionally allowing overridden TIdAuthentication.DoNext() methods to decide what to do with the values they have available at their first step.  If Next returns wnAskTheProgram, and OnProxyAuthorization is not assigned, bail out and stop the current request, same as if OnProxyAuthorization had returned Handled=False.

Similar logic could then be applied to TIdCustomHTTP.DoOnAuthorization() as well.

rlebeau:
Contact me privately, I'll send you my proposed changes to TIdHTTP.  I don't have access to any NTLM server that I can test with.

rlebeau:
Attached is a patch for `IdHTTP.pas` for my proposed changes. Before I consider checking them in, can you test it? I don't have access to any NTLM server that I can test with.
[IdHTTP patch.zip](https://github.com/IndySockets/Indy/files/2550620/IdHTTP.patch.zip)

markjforte:
Hi Remy, sorry, ran out of time today with other priorities.  I'll try to make time to do so tomorrow morning.

markjforte:
Hi Remy, I tested your changes and Yes, they work in our environment.  Note that I did not test thoroughly... I simply made a build with your changes (and without my changes) and verified that our proxy authentication occurred as it should, which it did.  I did no other testing.  Hope this helps.

PR #236, opened 2018-10-31
Fix for TIdSSPINTLMAuthentication

markjforte:
TIdSSPINTLMAuthentication currently does not work. This appears to have been an issue for quite some time as I found other folks that appear to have had the same issue way back in 2012:
http://codeverge.com/embarcadero.delphi.winsock/tidhttp-not-retrying-after-407-res/1073579
Here is another thread that seems to be related to the issue:
https://stackoverflow.com/questions/15111605/indy-tidhttp-integrated-security-authentication-issue-with-sspi
Note that there is an existing open issue, but I'm not sure if my fix addresses it, or if it is for something different?:
https://github.com/IndySockets/Indy/issues/87
Here is what I changed:
IdAuthentication.pas - Added virtual method TIdAuthentication.AuthorizationAttemptMerited.  Contains exact same logic that used to be TIdCustomHTTP.DoOnProxyAuthorization.
TIdHTTP.pas - Updated TIdCustomHTTP.DoOnProxyAuthorization to call TIdAuthentication.AuthorizationAttemptMerited rather than doing the logic itself.
IdAuthenticationSSPI.pas - TIdSSPINTLMAuthentication.AuthorizationAttemptMerited overrides default logic.  For this class authorization is always merited as no username or password is needed.

### DIFF
--- a/Lib/Protocols/IdAuthentication.pas
+++ b/Lib/Protocols/IdAuthentication.pas
@@ -85,6 +85,8 @@ type
 
     function Authentication: String; virtual; abstract;
     function KeepAlive: Boolean; virtual;
+    function AuthorizationAttemptMerited(
+      const AProxyPasswordeExists, AOnProxyAuthorizationExists: Boolean): Boolean; virtual;
     function Next: TIdAuthWhatsNext;
 
     property AuthParams: TIdHeaderList read FAuthParams write SetAuthParams;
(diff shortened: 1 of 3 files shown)


markjforte:
Hi Remy, note that I realize that my fix is not an ideal fix.  However, since I'm not at all familiar with this code I wanted to be very careful to not introduce any unanticipated side effects.  I believe that my solution should be very safe as I'm pretty sure that it shouldn't change the behavior of TIdCustomeHTTP or any of the IdAuthentication descendants (other than TIdSSPINTLMAuthentication).  No offense will be taken if you don't merge... Just wanted to put this out there in case others come across this issue.  Thanks!

rlebeau:
I'm not really liking the proposed approahc of adding a virtual method to TIdAuthentication.

But, it might be worth to instead tweak TIdCustomHTTP.DoOnProxyAuthorization() to simply remove the initial check for OnProxyAuthorization and ProxyParams.ProxyPassword being assigned, and let TIdHTTP call LAuth.Next unconditionally allowing overridden TIdAuthentication.DoNext() methods to decide what to do with the values they have available at their first step. If Next returns wnAskTheProgram, and OnProxyAuthorization is not assigned, bail out and stop the current request, same as if OnProxyAuthorization had returned Handled=False.

Similar logic could then be applied to TIdCustomHTTP.DoOnAuthorization() as well.

markjforte:
Sounds like a better solution, but I have nowhere near enough knowledge about this code (i.e. all the various TIdAuthentication descendants, etc) to opine if that would be safe or not.  I'll leave it to you.  Thanks!

markjforte:
Hi Remy, I assume you checked in changes to the SVN repository?  And if so, how long before they'll show up here on GitHub?  P.S. Thanks again for your dedication to this Indy library!

rlebeau:
Yes, it was checked in to SVN, Indy's GitHut repo is just a mirror for now.  It will hopefully be synced soon.  In the meantime, there is another repo (https://github.com/graemeg/indy) that is already synced.

Fulgan:
Pull requests are not possible at this time: this repo is only a mirror of the main SVN repo (http://ww2.indyproject.org/Sockets/Download/svn.EN.aspx)

PR #243, opened 2019-01-16
Update IdIPMCastClient.pas

dpcorbett:


rlebeau:
The fix has been checked in to Indy's SVN (rev 5493) and will sync with GitHub shortly.

Issue #218, opened 2018-07-03
Access Violation in SSL_accept() when deactivating TIdTCPServer


rlebeau:
A user is experiencing an access violation in `SSL_accept()` (see https://stackoverflow.com/questions/51150235/) when `TIdTCPServer` is setup to use OpenSSL and there is an open connection that has NOT yet negotiated an SSL/TLS session when `TIdTCPServer` is being deactivated.

This can be reproduced using the following code, with Putty making a RAW connection:

```
procedure TSslCrash.HandlerOnExecute(AContext: TIdContext);
begin
  //
end;

procedure TSslCrash.HandlerOnConnect(AContext: TIdContext);
begin
  TIdSSLIOHandlerSocketBase(AContext.Connection.IOHandler).PassThrough := False;
end;

procedure TSslCrash.ButtonStartClick(Sender: TObject);
begin
  LServer := TIdTCPServer.Create;
  LIOHandler := TIdServerIOHandlerSSLOpenSSL.Create;

  LIOHandler.SSLOptions.Mode := sslmServer;
  LIOHandler.SSLOptions.Method := sslvTLSv1_2;
  LIOHandler.SSLOptions.VerifyMode := [];
  LIOHandler.SSLOptions.VerifyDepth := 0;
  LIOHandler.SSLOptions.CertFile := 'localhost.crt';
  LIOHandler.SSLOptions.RootCertFile := 'localhost.crt';
  LIOHandler.SSLOptions.KeyFile := 'localhost.key';

  LServer.Bindings.Add.Port := 10000;
  LServer.IOHandler := LIOHandler;
  LServer.OnExecute := HandlerOnExecute;
  LServer.OnConnect := HandlerOnConnect;
  LServer.Active := True;

  //Now open a RAW connection with Putty on port 10000 and keep it open
end;

procedure TSslCrash.ButtonStopClick(Sender: TObject);
begin
  if Assigned(LServer) then begin
    LServer.Active := False;  //This causes an AV in TIdSSLSocket.Accept

    FreeAndNil(LIOHandler);
    FreeAndNil(LServer);
  end;
end;
```

rlebeau:
[...]

PR #270, opened 2019-11-20
Allow to terminate Recv/Send methods if fSSL was destroyed

evgeny-k:
I'm using `TIdHTTP`/`SSL` in background thread. 
if I call `Post` method, I have AV after calling `Disconnect` in main thread because `TIdSSLSocket` is destroyed but `TIdSSLSocket.Send` is still executed ....

call-stack (main thread) when TIdSSLSocket is freed:
>     IdSSLOpenSSL.TIdSSLSocket.Destroy
>     System.TObject.Free
>     IdSSLOpenSSL.TIdSSLIOHandlerSocketOpenSSL.Close
>     IdTCPConnection.TIdTCPConnection.Disconnect(???)
>     IdTCPConnection.TIdTCPConnection.Disconnect

AV in `TIdSSLSocket.Send` because `fSSL` is `nil`:

```
function TIdSSLSocket.Send(const ABuffer: TIdBytes; AOffset, ALength: Integer): Integer;
var
  ret, err: Integer;
begin
  Result := 0;
  repeat
    ret := SSL_write(fSSL, @ABuffer[AOffset], ALength); //<<<< AV
```
call-stack for this is

>     IdSSLOpenSSL.TIdSSLSocket.Send(...)
>     IdSSLOpenSSL.TIdSSLIOHandlerSocketOpenSSL.SendEnc(???,???,224)
>     IdSSL.TIdSSLIOHandlerSocketBase.WriteDataToTarget(???,???,224)
>     IdIOHandler.TIdIOHandler.WriteDirect(...)
>     IdIOHandler.TIdIOHandler.WriteBufferFlush(???)
>     IdIOHandler.TIdIOHandler.WriteBufferFlush
>     IdIOHandler.TIdIOHandler.WriteBufferClose
>     IdHTTP.TIdHTTPProtocol.BuildAndSendRequest(???)
>     IdHTTP.TIdCustomHTTP.ConnectToHost($7EA93560,???)
>     IdHTTP.TIdCustomHTTP.DoRequest(???,'https://.....',$7E91CBF0,$7E9B7AA0,(...))
>     IdHTTP.TIdCustomHTTP.Post('https://......',$8663B0,$7E9B7AA0)

if we add checking for `nil` for `fSSL` we can avoid AV

### DIFF
--- a/Lib/Protocols/IdSSLOpenSSL.pas
+++ b/Lib/Protocols/IdSSLOpenSSL.pas
@@ -3828,6 +3828,7 @@ var
   ret, err: Integer;
 begin
   repeat
+    if fSSL = nil then Break;  
     ret := SSL_read(fSSL, PByte(ABuffer), Length(ABuffer));
     if ret > 0 then begin
       Result := ret;
@@ -3852,6 +3853,7 @@ var
 begin
   Result := 0;
   repeat
+    if fSSL = nil then Break;
     ret := SSL_write(fSSL, @ABuffer[AOffset], ALength);
     if ret > 0 then begin
       Inc(Result, ret);


rlebeau:
Checking for nil is not good enough, all it does is introduce a new race condition and doesn't address the root issue of the TIdSSLSocket object being freed while it is still actively being used. This same issue also affects TIdTCPServer, see https://github.com/IndySockets/Indy/issues/218. In that ticket, just closing the underlying socket without destroying the SSL object worked. A similar solution would likely apply here, too.

evgeny-k:
it doesn't work for me. 
I'm using client-side component (`TIdHTTP`). 
in my case `IdHTTP.Socket.Binding` is nil so I can't call `IdHTTP.Socket.Binding.CloseSocket()`
If I call `IdHTTP.Socket.Close;` - it calls 
```
procedure TIdSSLIOHandlerSocketOpenSSL.Close;
begin
  FreeAndNil(fSSLSocket);
  if not IsPeer then begin
    FreeAndNil(fSSLContext);
  end;
  inherited Close;
end;
```
so `fSSLSocket` is destroyed and AV is raised again.

We can add some variable like `fTerminated` that would be initialized in `destructor` and check for it in `Recv`/`Send` methods. at least it will allow to terminate SSL methods ...

rlebeau:
> in my case `IdHTTP.Socket.Binding` is nil so I can't call `IdHTTP.Socket.Binding.CloseSocket()`

If the `Binding` is nil then there is no socket allocated at all (either the IOHandler was never opened, or it has already been destroyed), so there should be no SSL object allocated, either.

> If I call `IdHTTP.Socket.Close;` - it calls
> 
> ```
> procedure TIdSSLIOHandlerSocketOpenSSL.Close;
> begin
>   FreeAndNil(fSSLSocket);
>   if not IsPeer then begin
>     FreeAndNil(fSSLContext);
>   end;
>   inherited Close;
> end;
> ```

Yes, I'm aware of that.  `Close()` freeing the `fSSLSocket` object is the root of the [other ticket](https://github.com/IndySockets/Indy/issues/218) I linked to earlier.  Same issue here.  So my suggestion is to NOT free that object anymore in `Close()`, just shutdown+close (not deallocate) the underlying socket instead.  Free the `fSSLSocket` in the destructor instead.

> We can add some variable like `fTerminated` that would be initialized in `destructor` and check for it in `Recv`/`Send` methods. at least it will allow to terminate SSL methods ...

That would not work.  If the `Recv`/`Send` methods are called on an invalid object, there would be no valid `fTerminated` member available to access.  Undefined behavior.

Fulgan:
I agree that an AV creates an unnecessary security risk in the library. However, I also agree that this AV is a consequence of anyother problem in the software itself.
Therefore, I suggest this pull request be changed to raise an exception if the fSSL reference is nil. That will prevent the AV without supressing the error.

rlebeau:
There is no **security** risk in an AV being raised.  And there is really no difference in raising one type of exception versus another.  I don't think checking for a nil SSL object is the right solution in this situation.  It doesn't actually solve anything.

Issue #274, opened 2019-12-09
Lazarus compile errors when using FPC 3.0.4a (last and most common stable)


TommySlokky:
After having updated to Catalina + newest Cocoa source, I updated Indy as well (due to strange new errors with TSLV1 I did not experience on Mojave, so wondering if something changed regarding that in Catalina besides notarization/entitlements/symlink issues... but that is for another post if problems persists updating and researching further) and encountered some compile time errors inside IdGlobal.pas ...

Quote
IdGlobal.pas(4607,57) Error: Typecast has different size (4 -> 8 ) in assignment
at

Code: Pascal [Select]
{$IFDEF THANDLE_64}
Result := THandle(InterlockedExchange64(Int64(VTarget), Int64(AValue)));
{$ENDIF}

and two of this

Quote
IdGlobal.pas(8523,17) Error: Identifier not found "PRawByteString"
at

Code: Pascal [Select]
SetCodePage(PRawByteString(@Result)^, GetEncodingCodePage(ADestEncoding), False);

...

When I run "svn up" on my fprsrc directory it says 43661

---Speculation--

I believe this is because Indy assumes one is using FPC trunk? (where PRawByteString is defined) And not the stable 3.0.4a release. I am a bit hesitant upgrading FPC to a non-stable release...

The PRawByteString is an easy "hotfix" oneself, but if Indy now assumes {$IFDEF THANDLE_64}
(maybe this changed in FPC 3.2.x / 3.3.x ?)

I guess that could leave to unforseen problems (by me) if not "hotfixed" correctly. Assuming this is a wontfix but I am not prepared to upgrade non-stable FPC, do you have recommendation if it is necessary to fix the define itself? Or just this one place in code?

Bi0T1N:
I assume you've a 64bit CPU while your FPC version is 32bit. Is this correct? Maybe it's an issue with old FPC + Catalina?
Please post the output from `fpc -vc`.

TommySlokky:
Been generating 64bit executables for a couple of years with my setup. (Otherwise I would not even be able to run my apps on Catalina + not get warnings with my release builds)

However, I found many posts on Google where it is explicitly states THandle on 64bit FPC is still only 4 bytes on purpose. However, those posts are rather old, so I am not sure it is fully uptodate.

Bi0T1N:
> 
> > That currently breaks with FPC64 on Mac, because THandle is not 64-bit.
> 
[...]

PR #275, opened 2019-12-10
Define PRawByteString for older FPC

Bi0T1N:
added in FPC svn 39662

fixes one issue from #274

### DIFF
--- a/Lib/Core/IdCompilerDefines.inc
+++ b/Lib/Core/IdCompilerDefines.inc
@@ -1443,6 +1443,7 @@
   {$ENDIF}
   {$IFDEF FPC_3_1_1_OR_ABOVE}
     {$DEFINE HAS_STATIC_TThread_ForceQueue} // requires rev 37359+
+    {$DEFINE HAS_PRawByteString}
   {$ENDIF}
 {$ENDIF}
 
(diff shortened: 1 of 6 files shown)


rlebeau:
When defining `PRawByteString` in `IdGlobal.pas`, please wrap it in an `{$IFDEF SetCodePage}` block:
```
  {$IFDEF HAS_SetCodePage}
    {$IFNDEF HAS_PRawByteString}
  PRawByteString = ^RawByteString;
    {$ENDIF}
  {$ENDIF}
````
The only time that Indy uses `RawByteString` is when calling `SetCodePage()`, which is not available in Delphi 2007 and earlier.  Your proposed change will not compile in those older versions of Delphi because `RawByteString` did not exist yet.

rlebeau:
`SetCodePage()` and `RawByteString` have been [available in FPC since 3.0](https://wiki.freepascal.org/Unicode_Support_in_Lazarus), but your proposal defines `HAS_PRawByteString` for FPC 3.1.1+ instead. Are you sure that FPC 3.0 did not introduce `PRawByteString` when it introduced `RawByteString`?

Bi0T1N:
1. Added `{$IFDEF HAS_SetCodePage}` around the block. Indy supports to many different compiler versions... :D
2. Yes, I checked it before writing the patch. It was added in svn 39662 (https://github.com/graemeg/freepascal/commit/b3708d24aded4470af5396eceed255b0892c7803) but never made it into 3.0.4 (https://svn.freepascal.org/svn/fpc/tags/release_3_0_4/rtl/inc/systemh.inc) nor in 3.0.4a for MacOS (https://svn.freepascal.org/svn/fpc/tags/release_3_0_4a_macos_10_14/rtl/inc/systemh.inc)

rlebeau:
I merged the pull request, and then added a small tweak of my own on top of it.

PR #281, opened 2020-01-31
Define global TLSVersion in IdSSL

Bi0T1N:
this allows to have one set of TLS versions for all SSL/TLS IO handler

### DIFF
--- a/Lib/Protocols/IdSSL.pas
+++ b/Lib/Protocols/IdSSL.pas
@@ -117,8 +117,8 @@ type
   TIdServerSSLClass = class of TIdServerIOHandlerSSLBase;
 
 type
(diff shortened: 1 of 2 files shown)


Bi0T1N:
I've reverted the changes and called the type TLS versions as this is more appropriate.

Bi0T1N:
Firstly I named it `tlsvTLS_method` but to know the meaning you need to read the documentation for it. So I thought naming it `tlsvTLS_flexible` (or `tlsvTLS_highest`) is more intuitive as it negotiates the highest possible version between client and server (therefore flexible).
The define is also not meant for use with OpenSSL 1.0.2 (it's EOL and should therefore be ignored for new code) and almost all libraries removed SSLv3 a long time ago. No need to keep it in the possible versions. 
I've created the types so that all new SSL IO handler can make use of it to avoid having different name schemes for different IO handler. (OpenSSL, LibreSSL, LibTLS, ...)

PR #282, opened 2020-02-01
Add alias for C size_t type and fix wrong definition of TIdC_UINT32

Bi0T1N:
This MR fixes the wrong definition of `TIdC_UINT32` (signed integer instead of unsigned) and also adds aliases for `size_t`, `ssize_t` and `time_t` which is useful for API header translation.

### DIFF
--- a/Lib/Core/IdCompilerDefines.inc
+++ b/Lib/Core/IdCompilerDefines.inc
@@ -1376,10 +1376,14 @@
     {$DEFINE HAS_SharedSuffix}
   {$ENDIF}
   {$IFDEF FPC_2_2_4_OR_ABOVE}
-    // size_t and psize_t are only available on Unix systems (FreeBSD, Linux, etc)
+    // these types are only available on Unix systems (FreeBSD, Linux, etc)
     {$IFDEF UNIX}
       {$DEFINE HAS_SIZE_T}
       {$DEFINE HAS_PSIZE_T}
+      {$DEFINE HAS_SSIZE_T}
+      {$DEFINE HAS_PSSIZE_T}
+      {$DEFINE HAS_TIME_T}
+      {$DEFINE HAS_PTIME_T}
     {$ENDIF}
   {$ENDIF}
   {$DEFINE HAS_PtrInt}
--- a/Lib/FCL/IdCompilerDefines.inc
+++ b/Lib/FCL/IdCompilerDefines.inc
@@ -1376,10 +1376,14 @@
     {$DEFINE HAS_SharedSuffix}
   {$ENDIF}
   {$IFDEF FPC_2_2_4_OR_ABOVE}
-    // size_t and psize_t are only available on Unix systems (FreeBSD, Linux, etc)
+    // these types are only available on Unix systems (FreeBSD, Linux, etc)
     {$IFDEF UNIX}
       {$DEFINE HAS_SIZE_T}
       {$DEFINE HAS_PSIZE_T}
+      {$DEFINE HAS_SSIZE_T}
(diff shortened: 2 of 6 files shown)


Bi0T1N:
The two types are added for FreePascal as well. It's defined outside of all `IFDEF` blocks.
The types `csize_t` and `pcsize_t` are also just mappings to `cuint64` or `cuint32`.
If you prefer to use them, I can make the changes but I don't know what you want to use for Delphi (to be working in all ancient versions) then.

rlebeau:
> The two types are added for FreePascal as well. It's defined outside of all IFDEF blocks.

I would prefer to keep the existing coding style of keeping all the FPC defines separate from the Delphi defines.  Define TIdC_SIZET/TIdC_SIZET in the FPC block using csize_t/cssize_t, and defined separately in the Delphi block using Native(U)Int or TIdC_(U)INT32/64.  Something like this, maybe:

```
type 
  {$IFDEF FPC}
  ...
  TIdC_SIZET = csize_t;
  TIdC_PSIZET = pcsize_t;
  TIdC_SSIZET = cssize_t;
  TIdC_PSSIZET = pcssize_t;

  {$ELSE}
  ...

  {$IFDEF HAS_SIZE_T}
  TIdC_SIZET = size_t;
  {$ELSE}
    {$IFDEF HAS_NativeUInt}
  TIdC_SIZET = NativeUInt;
    {$ELSE}
      {$IFDEF CPU32}
  TIdC_SIZET = TIdC_UINT32;
      {$ENDIF}
      {$IFDEF CPU64}
  TIdC_SIZET = TIdC_UINT64;
      {$ENDIF}
    {$ENDIF}
  {$ENDIF}
  TIdC_PSIZET = ^TIdC_SIZET;

  {$IFDEF HAS_SSIZE_T}
  TIdC_SSIZET = ssize_t;
  {$ELSE}
    {$IFDEF HAS_NativeInt}
  TIdC_SSIZET = NativeInt;
    {$ELSE}
      {$IFDEF CPU32}
  TIdC_SSIZET = TIdC_INT32;
      {$ENDIF}
      {$IFDEF CPU64}
  TIdC_SSIZET = TIdC_INT64;
      {$ENDIF}
    {$ENDIF}
  {$ENDIF}
  TIdC_PSSIZET = ^TIdC_SSIZET;

  ...

  {$ENDIF}
```

Bi0T1N:
Using `cssize_t` with FPC 3.3.1 on Windows fails:
```
Error: Identifier not found "cssize_t"
Error: Error in type definition
Error: Identifier not found "pcssize_t"
Error: Error in type definition
```
So it's highly likely that it also fails with older versions of FPC on non-unix? systems.

Can't find anything about `ssize_t` in Delphi at all, so probably not available.

rlebeau:
FreePascal has ssize_t in the BaseUnix and UnixType units.

Bi0T1N:
> 
> 
> FreePascal has ssize_t in the BaseUnix and UnixType units.

```
Fatal: Can't find unit BaseUnix used by IdCTypes
Fatal: Can't find unit UnixType used by IdCTypes
```
These units are not available on Windows platforms!
Or do you want `IFDEFs` for FPC which uses `ssize_t` on unix platforms?

rlebeau:
> These units are not available on Windows platforms!

I never said they were.

> Or do you want `IFDEFs` for FPC which uses `ssize_t` on unix platforms?

Yes (Indy already has a HAS_SIZE_T conditional in IdCompilerDefines.inc).  And then uses appropriate alternative types on Windows.

Bi0T1N:
Made the changes as requested.
Additionally added `time_t` which is also used in C headers.

PR #293, opened 2020-03-17
Removed empty unit

mezen:
Small code clean ups to remove unused GSSLRegistry

and the empty unit IdServerIOHandlerSSLOpenSSL.


JNPSoftware:
Maybe it's time to remove .NET too?

mezen:
Even if external tools make use of it, they could still create an own list and register the `SLLIOHandler`.

rlebeau:
Support for .NET and pre-Unicode compilers are already slated to be removed in Indy 11.

Issue #300, opened 2020-05-15
TIdFTP fails on Active mode transfer connection with vsFTPd


rlebeau:
A user [reported on StackOverflow](https://stackoverflow.com/questions/61822360/) that `TIdFTP` is not accepting an *Active* mode file transfer from vsFTPd.

According to the below Wireshark capture, vsFTPd is creating the transfer connection *BEFORE* sending a response to the `RETR` command:

![Pyjg1](https://user-images.githubusercontent.com/15006977/82082379-c442d400-969c-11ea-89d8-253232fa85cb.png)

The user also tested with ICS and was able to perform the same transfer OK:

![ieIwm](https://user-images.githubusercontent.com/15006977/82083561-cdcd3b80-969e-11ea-84b7-c8747359621b.png)

ICS's `TFtpClient` allows that connection before receiving the `RETR` response.  But Indy's `TIdFTP` waits for the `RETR` response to arrive before it will then accept the transfer connection (this also applies to `STOR`/`STOU`/`APPE` requests, too), due to its use of `TIdTCPConnection.SendCmd()`.

Re-reading [RFC 959](https://tools.ietf.org/html/rfc959), it says the following:

> The passive data transfer process (this may be a user-DTP or a second server-DTP) shall "listen" on the data port prior to sending a transfer request command.  The FTP request command determines the direction of the data transfer.  **The server, upon receiving the transfer request, will initiate the data connection to the port.  When the connection is established, the data transfer begins between DTP's, and the server-PI sends a confirming reply to the user-PI**.

ICS is asynchronous, so this situation is not a big deal for it to handle.  But Indy uses blocking sockets, so `TIdFTP` will have to be updated to account for this situation, likely by monitoring both command and transfer ports simultaneously so it can act accordingly regardless of the order in which the transfer connection and the command response arrive in.

PR #301, opened 2020-05-16
Fix for active mode transfers in IdFTP.pas

rlebeau:
Fix for #300. Updating TIdFTP.InternalGet() and TIdFTP.InternalPut() to account for FTP servers that open the active mode connection before sending a 1xx response.


rlebeau:
Confirmed working by the user who reported the issue on StackOverflow.

PR #412, opened 2022-05-15
Readme.md + 1 png about Indy components

PizzaProgram:


rlebeau:
I appreciate the contribution. But, why are you advertising Indy in a 20-year-old compiler? And the wording of the readme is a bit misleading/inaccurate. And the statement made in the DLL section is not very kind.  Also, the image link is referring to the PNG using the wrong path.

This needs to be re-written before it can be merged.

PizzaProgram (commented):
Fixed.  
Added +1 screenshot,  
also HLP files.

PizzaProgram:
Hi,
-

 I'm still learning how to properly use Github...

  I have an XE8 installed on my other PC, I added a 2th pic too, to show both old and new worlds.  
  
IMHO it is important to show a Delphi-7 picture, because many people still use D7, but some of them do NOT know that they should upgrade Indy inside it!  

> And the statement made in the DLL section is not very kind

 **Sorry**, it mend to be funny!   
But also it would be nice if someone would change that part of the text and explain, what version should be used.  
I will change the text right now!  


# Can You please explain it to me?
 ... or should I leave that part empty, and you will re-edit the readme.md ?
 - what version of DLL is good for the current branch?
 - which one for the 1.1.1 ?

PizzaProgram:
Is it OK this way?

PR #414, opened 2022-07-14
add a README

corneliusdavid:


README.md
```diff
@@ -0,0 +1,6 @@
+# Indy - Internet Direct
+
+Indy is a well-known internet component suite for **Delphi**, **C++Builder**, and **Free Pascal** providing both low-level 
+support (TCP, UDP, raw sockets) and over a 120 higher level protocols (SMTP, POP3, NNT, HTTP, FTP) for building both client and server applications.
+
+For instructions on upgrading the default installed version of Indy in your IDE and for links to the documentation, visit the Wiki.
```
PatrickvL, line 6: What wiki? At least mention an URL!

corneliusdavid: I didn't at first because I had problems with relative paths; but thinking about it again, it shouldn't be a relative path, it should be an explicit one! I've added a commit--thanks for prompting me!

rlebeau (approved):

PR #418, opened 2022-07-23
batch scripts to remove default Indy library files in Delphi XE3-11

corneliusdavid:
I've had to remove/upgrade Indy in multiple versions of Delphi so I wrote some batch files because there's a lot of files and I don't want to manually do this every time. As I thought about making them public, I added logging, warning messages, and used the `computil` utility already in the Indy 10 library to set up the environment. Then I went one step further and put all the delete commands in a separate batch file and wrote smaller ones that called it for the various versions of Delphi. This reduced copy/paste errors and made it easier to change a delete command or warning message for all versions of Delphi.

I wrote the scripts for the most recent versions first, doing a lot of testing in 10.4 and 11. These versions are the ones I use most and have the most platforms and files. After those were done, I copied the scripts for older versions. They have been tested in Delphi 11 Alexandria, 10.4 Sydney, 10.1 Berlin, and XE3-XE8.

I hope these are useful for others--and I hope we can manage packages differently in Indy 11!


corneliusdavid:
After discussing this offline, I'm going to streamline these batch files and submit a different PR.

ThomasLevering:
if you add this at the beginning of CleanDxx.cmd:

cd "%~dp0"

then left Click Execute as Admin is working, without cmd.exe

corneliusdavid:
@ThomasLevering , I don't want the batch scripts to leave the user in a different folder from where they started. I suppose I could use `pushd` and `popd` though. I'll have to try this out. Thanks for the suggestion.

PR #426, opened 2022-08-23
Fix BIO_ctrl parg argument in BIO_set_nbio_accept

iadcode:
**Problem**
- There is a direct Char -> PChar cast in `IdSSLOpenSSLHeaders.BIO_set_nbio_accept`.

**Background**
Casting a Character type to a Character Pointer type is non-portable between different versions of Delphi.

In Delphi 11, there's a significant breaking change in the code generation around Char/AnsiChar -> PChar/PAnsiChar casts.
Rather than taking the ordinal value of the character and converting it to a pointer, a new single-character string is constructed and the pointer to that string is returned instead.

See:
- [Wrong results in cast from PWideChar to Char](https://quality.embarcadero.com/browse/RSP-29639)
- [Bug in PWideChar](https://quality.embarcadero.com/browse/RSP-29772)
- [Bad codegen breaks Winapi.Windows.CharUpperW()](https://quality.embarcadero.com/browse/RSP-31498)

### DIFF
--- a/Lib/Protocols/IdSSLOpenSSLHeaders.pas
+++ b/Lib/Protocols/IdSSLOpenSSLHeaders.pas
@@ -25481,9 +25481,11 @@ end;
 //* #define BIO_set_nbio(b,n)	BIO_ctrl(b,BIO_C_SET_NBIO,(n),NULL) */
 function BIO_set_nbio_accept(b : PBIO; n : TIdC_INT) : TIdC_LONG;
 {$IFDEF USE_INLINE} inline; {$ENDIF}
+const
+  a: array[0..1] of TIdAnsiChar = (TIdAnsiChar('a'), TIdAnsiChar(#0));
 begin
   if n <> 0 then begin
-    Result := BIO_ctrl(b, BIO_C_SET_ACCEPT, 1, PIdAnsiChar('a'));
+    Result := BIO_ctrl(b, BIO_C_SET_ACCEPT, 1, @a[0]);
   end else begin
     Result := BIO_ctrl(b, BIO_C_SET_ACCEPT, 1, nil);
   end;


rlebeau:
Your proposed change is incorrect.  The input value that `BIO_set_nbio_accept()` is expected to pass to `BIO_ctrl()` is _supposed_ to be a pointer to a null-terminated `"a"` character string, not a single character that has been casted to a pointer.  The original code should work just fine in all Delphi versions.  But, if for some reason it is not, then the _proper_ solution would be to first assign the `'a'` character to a local `TIdAnsiString` or `array[0..1] or TIdAnsiChar`, and then pass a pointer to that to `BIO_ctrl()`.

iadcode:
First off, thanks for the prompt response.

>  The input value that BIO_set_nbio_accept() is expected to pass to BIO_ctrl() is supposed to be a pointer to a null-terminated "a" character string

Oops. I made an assumption here based on the existing behaviour in Indy on Delphi compilers prior to Alexandra, which will not give you a pointer to a null-terminated character string.

> BTW, none of your QP links are working. They all report a "Permission violation" error.

Ah, you have to log in to view them. The linked issues just provide some background/context for the codegen change.

> The original code should work just fine in all Delphi versions

On every version prior to Delphi 11, the value of `PChar('a')` is the ordinal value of the character 'a'. From Delphi 11 onward, the value of `PChar('a')` is an actual pointer to a temporary string containing 'a'.

```
procedure Test;
var
  P: PAnsiChar;
  N: NativeInt;
begin
  P := PAnsiChar('a');
  N := NativeInt(P);

{$if CompilerVersion < 35.0}
  // In every version of Delphi up until Alexandria...
  // Ordinal value of 'a'
  Assert(N = 97);
{$else}
  // From Alexandria onwards...
  // Pointer to a temporary string containing 'a'
  Assert(N <> 97);
{$endif}
end;
```

This would seem to indicate that the existing behaviour is what's broken, and it actually only started to work as expected in Alexandria.

I've updated my commit with one of your suggested solutions, what do you think?

TommiPrami:
> And BTW, none of your QP links are working. They all report a "Permission violation" error.

Embarcadero had some problems, couple days ago all QPs started to give me that werror, that has happened couple times before,

Some kind of configuration glitch or they are doing some maintenance or so., Hard to tell....

-Tee-

rlebeau:
> Oops. I made an assumption here based on the existing behaviour in Indy on Delphi compilers prior to Alexandra, which will not give you a pointer to a null-terminated character string.

And I was under the assumption that the existing behavior would always produce a null-terminated string.  Guess not.

> Ah, you have to log in to view them.

I was logged in at the time, and could not see them, even though they are in the public RSP project.  But I just tried again and now I can see them.  Funny, one of them (RSP-31498) is my own report.

> On every version prior to Delphi 11, the value of `PAnsiChar('a')` is the ordinal value of the character 'a'. From Delphi 11 onward, the value of `PAnsiChar('a')` is an actual pointer to a temporary string containing 'a'.

The latter is the desired behavior in this case.  But, best not to rely on this anymore, I suppose.

> I've updated my commit with one of your suggested solutions, what do you think?

It should be fine. I will merge it.

rlebeau (approved):

GabrielNoga:
Stopped working with Delphi XE2 version.

I got the follow error when I run the Fullc_XE2.bat file:

`"C:\git\Indy\C16\IndyProtocols160.dproj" (Rebuild target) (1) ->
(_PasCoreCompile target) ->
  DCC : IdSSLOpenSSLHeaders.pas(25488) error : E2441 Inline function declared in interface section must not use local s
ymbol 'a'`

PR #429, opened 2022-08-26
Allow BIO_set_nbio_accept to be inlined again

### LINKED ISSUE #428: fullc_xe.bat fails to compile: E2441 Inline function declared in interface section must not use local symbol 'a'
Following the instructions of Updating Indy, I got to the point where I need to run batch file fullc_xe.bat. That fails with following message:

IdDeprecatedImplBugOff.inc(6)
IdDeprecatedImplBugOn.inc(10)
IdDeprecatedImplBugOff.inc(6)
IdDeprecatedImplBugOn.inc(10)
IdDeprecatedImplBugOff.inc(6)
IdDeprecatedImplBugOn.inc(10)
IdDeprecatedImplBugOff.inc(6)
IdDeprecatedImplBugOn.inc(10)
IdGlobalProtocols.pas(5377)
IdDeprecatedImplBugOff.inc(6)
IdDeprecatedImplBugOn.inc(10)
IdHeaderList.pas(507)
IdCompilerDefines.inc(2018)
IdCompilerDefines.inc(2018)
IdCompilerDefines.inc(2018)
IdCoder.pas(481)
IdCoder3to4.pas(408)
IdCoderMIME.pas(188)
IdAuthentication.pas(416)
IdResourceStringsOpenSSL.pas(36)
IdCompilerDefines.inc(2018)
IdCompilerDefines.inc(2018)
IdFIPS.pas(230)
**IdSSLOpenSSLHeaders.pas(25488) Error: E2441 Inline function declared in interface section must not use local symbol 'a'**
IdSSLOpenSSLHeaders.pas(26993)
IdAuthenticationNTLM.pas(2102) Fatal: F2063 Could not compile used unit 'IdSSLOpenSSLHeaders.pas'

I downloaded the master branch on Aug 25 2022. The last merged pull request at that time was: [Merge pull request](https://github.com/IndySockets/Indy/commit/e8742696acf4076b6b4c2fe35c9f35c0ff7fe54b) https://github.com/IndySockets/Indy/pull/426 [from iadcode/char_to_pchar_c](https://github.com/IndySockets/Indy/commit/e8742696acf4076b6b4c2fe35c9f35c0ff7fe54b)

iadcode:
It turns out that `const` declared within a function counts as a "local symbol" for the purposes of inlining and fails the build.

### DIFF
--- a/Lib/Protocols/IdSSLOpenSSLHeaders.pas
+++ b/Lib/Protocols/IdSSLOpenSSLHeaders.pas
@@ -25480,10 +25480,13 @@ end;
 
 //* #define BIO_set_nbio(b,n)	BIO_ctrl(b,BIO_C_SET_NBIO,(n),NULL) */
 function BIO_set_nbio_accept(b : PBIO; n : TIdC_INT) : TIdC_LONG;
-const
-  a: array[0..1] of TIdAnsiChar = (TIdAnsiChar('a'), TIdAnsiChar(#0));
+{$IFDEF USE_INLINE} inline; {$ENDIF}
+var
+  a: array[0..1] of TIdAnsiChar;
 begin
   if n <> 0 then begin
+    a[0] := TIdAnsiChar('a');
+    a[1] := TIdAnsiChar(#0);
     Result := BIO_ctrl(b, BIO_C_SET_ACCEPT, 1, @a[0]);
   end else begin
     Result := BIO_ctrl(b, BIO_C_SET_ACCEPT, 1, nil);


rlebeau:
Sorry, I hadn't seen this pull request yet when I checked in my own fix for the issue.  I just removed the `inline` rather than changing the `const` into a `var`.

rlebeau (changes requested):
Please resync with the latest master, and either re-submit or close this PR as needed.

iadcode:
Hi @rlebeau, I've resynced with latest master and renamed the commit/PR appropriately.

rlebeau:
Rather than using a local `var`, if the `const` were moved outside of the function, would that allow the function to be inlined?

Cirras:
> error E2441: Inline function declared in interface section must not use local symbol 'a'

Afraid not, extracting the const gives the same error as before.

iadcode:
> error E2441: Inline function declared in interface section must not use local symbol 'a'

Afraid not, extracting the const gives the same error as before.

PatrickvL:
Would passing in `PAnsiChar('a'#0)` as an argument do the trick as well? That way no `a` variable is needed at all.

iadcode:
> Would passing in `PAnsiChar('a'#0)` as an argument do the trick as well? That way no `a` variable is needed at all.

`PAnsiChar(String('a'))` would probably work similarly, but I haven't checked.
I opted for the very explicit array of `PAnsiChar` because it seemed a sure thing that it would work consistently across different compilers/versions. After all, the original bug was basically a portability issue between Delphi compiler versions.

rlebeau:
> `PAnsiChar(String('a'))` would probably work similarly, but I haven't checked.

It would have to be `PAnsiChar(AnsiString('a'))` instead, if `PAnsiChar('a'#0)` does not work.

Cirras:
> > `PAnsiChar(String('a'))` would probably work similarly, but I haven't checked.
> 
> It would have to be `PAnsiChar(AnsiString('a'))` instead, if `PAnsiChar('a'#0)` does not work.

Since Indy supports all of these different compilers and compiler versions, I would expect the behavior here to be informed by the type mappings.

- `PIdAnsiChar` sometimes aliasesto `PAnsiChar` sometimes to `PByte`, and sometimes to `MarshaledAString`.
- There's no `TIdAnsiString` alias.

Would there be any consistent way to do the equivalent of `PAnsiChar(AnsiString('a'))`, or even `PAnsiChar('a'#0)`, then?

iadcode:
> > `PAnsiChar(String('a'))` would probably work similarly, but I haven't checked.
> 
> It would have to be `PAnsiChar(AnsiString('a'))` instead, if `PAnsiChar('a'#0)` does not work.

Since Indy supports all of these different compilers and compiler versions, I would expect the behavior here to be informed by the type mappings.

- `PIdAnsiChar` sometimes aliasesto `PAnsiChar` sometimes to `PByte`, and sometimes to `MarshaledAString`.
- There's no `TIdAnsiString` alias.

Would there be any consistent way to do the equivalent of `PAnsiChar(AnsiString('a'))`, or even `PAnsiChar('a'#0)`, then?

rlebeau:
> Would there be any consistent way to do the equivalent of `PAnsiChar(AnsiString('a'))`, or even `PAnsiChar('a'#0)`, then?

Not cleanly, it would probably end up looking something like this:

```
{$IFDEF USE_MARSHALLED_PTRS}
var
  M: TMarshaller;
{$ENDIF}
...
Result := BIO_ctrl(b, BIO_C_SET_ACCEPT, 1,
  {$IFDEF USE_MARSHALLED_PTRS}
  M.AsAnsi('a').ToPointer
  {$ELSE}
    PAnsiChar(
      {$IFDEF STRING_IS_ANSI}
      'a'#0
      {$ELSE}
      AnsiString('a')
      {$ENDIF}
    )
  {$ENDIF}
);
```

rlebeau (approved):

PR #442, opened 2022-12-24
idHl7 update to last release from original author for indy 10 and som…

shoraitek:
idHl7 update to last release from original author for indy 10 and some code fixes

  Rev 2.0   23/12/2022 19:52 PM EJPretorius / ShoraiTek
  Combined Indy code with last source code release by original author (Grahame Grieve) in 2013
    http://www.healthintersections.com.au/?p=1596
 Some extra changes around Keep Alive (to work as other units in indy project) made and logic review in some function between the
  two merge codesets

Will add Examle Client and Servers in the future for the Protocol


Lib/Protocols/IdHL7.pas
```diff
@@ -184,25 +186,30 @@   EHL7CommunicationError = class(EIdException)
...
-  TMessageArriveEvent = procedure(ASender: TObject; AConnection: TIdTCPConnection; AMsg: String) of object;
-  TMessageReceiveEvent = procedure(ASender: TObject; AConnection: TIdTCPConnection; AMsg: String; var VHandled: Boolean; var VReply: String) of object;
-  TReceiveErrorEvent = procedure(ASender: TObject; AConnection: TIdTCPConnection; AMsg: String; AException: Exception; var VReply: String; var VDropConnection: Boolean) of object;
+  TMessageArriveEvent = Procedure(ASender: TObject; AConnection: TIdTCPConnection; AMsg: AnsiString) Of Object;
+  TMessageReceiveEvent = Procedure(ASender: TObject; AConnection: TIdTCPConnection; AMsg: AnsiString; Var VHandled: Boolean; Var VReply: AnsiString) Of Object;
+  TReceiveErrorEvent = Procedure(ASender: TObject; AConnection: TIdTCPConnection; AMsg: AnsiString; AException: Exception; Var VReply: AnsiString; Var VDropConnection: Boolean) Of Object;
 
   TIdHL7 = class;
   TIdHL7ConnCountEvent = procedure (ASender : TIdHL7; AConnCount : integer) of object;
+  TIdPeerThread = TIdContext;
```
rlebeau, line 204: TIdPeerThread was an Indy 9 class, I'm not re-introducing it in Indy 10.  or Indy 10, please use TIdContext directly in methods as needed.

Lib/Protocols/IdHL7.pas
```diff
@@ -225,13 +235,16 @@   TIdHL7 = class(TIdBaseComponent)
...
     FIPMask: String;
     FIPRestriction: String;
+    FIPMaskVal : Cardinal;
+    FIPRestrictionVal : Cardinal;
     FIsListener: Boolean;
     FObject: TObject;
     FPreStopped: Boolean;
     FPort: Word;
-    FReconnectDelay: LongWord;
+    FReconnectDelay: Cardinal;
```
rlebeau, line 244: Why change a bunch of members from LongWord to Cardinal?  Please leave them as LongWord, or use UInt32 if you want to force a 32bit type.

Lib/Protocols/IdHL7.pas
```diff
@@ -278,17 +293,18 @@   TIdHL7 = class(TIdBaseComponent)
     procedure StartServer;
     procedure StopServer;
     procedure DropServerConnection;
-    procedure ServerConnect(AContext: TIdContext);
-    procedure ServerExecute(AContext: TIdContext);
-    procedure ServerDisconnect(AContext: TIdContext);
+    Procedure ServerConnect(AThread: TIdPeerThread);
+    Procedure ServerExecute(AThread: TIdPeerThread);
+    Procedure ServerDisconnect(AThread: TIdPeerThread);
```
rlebeau, line 298: Please don't use TIdPeerThread here, use TIdContext directly.

Lib/Protocols/IdHL7.pas
```diff
@@ -278,17 +293,18 @@   TIdHL7 = class(TIdBaseComponent)
...
 
     procedure CheckClientParameters;
     procedure StartClient;
     procedure StopClient;
     procedure DropClientConnection;
+    Procedure ReConnectFromTimeout;
 
-    procedure HandleIncoming(const AMsg : String; AConnection: TIdTCPConnection);
-    function HandleMessage(const AMsg: String; AConn: TIdTCPConnection; var VReply: String): Boolean;
+    Procedure HandleIncoming(Var VBuffer: AnsiString; AConnection: TIdTCPConnection);
```
rlebeau, line 306: Why change the AMsg parameter from const to var?

Lib/Protocols/IdHL7.pas
```diff
@@ -315,33 +331,39 @@   TIdHL7 = class(TIdBaseComponent)
...
   Published
     // basic properties
     property Address: String Read FAddress Write SetAddress;  // leave blank and we will be server
     property Port: Word Read FPort Write SetPort Default DEFAULT_PORT;
 
+    Property KeepAlive : Boolean read FKeepAlive write SetKeepAlive;
+
     // milliseconds - message timeout - how long we wait for other system to reply
-    property TimeOut: LongWord Read FTimeOut Write SetTimeOut Default DEFAULT_TIMEOUT;
+    Property TimeOut: Cardinal Read FTimeOut Write SetTimeOut Default DEFAULT_TIMEOUT;
```
rlebeau, line 364: I don't see the need to change the type in these properties.

Lib/Protocols/IdHL7.pas
```diff
@@ -556,6 +594,38 @@ destructor TIdHL7.Destroy;
   end;
 end;
 
+function TIdHL7.ConvertIPtoCardinal(const AStr: AnsiString): Cardinal;
```
rlebeau, line 597: Is this method really necessary?  Indy already has other functions for parsing IP addresses into numeric values. For example, `IdGlobal.IPv4ToUInt32()`

Lib/Protocols/IdHL7.pas
```diff
@@ -887,6 +1024,12 @@ procedure TIdHL7.StartServer;
     FServer.OnExecute := ServerExecute;
     FServer.OnDisconnect := ServerDisconnect;
     FServer.Active := True;
+    if FKeepAlive Then
+    Begin
+      d := $FFFFFFFF;
+      for i := 0 to FServer.Bindings.count - 1 Do
+        FServer.Bindings[i].SetSockOpt(Id_SOL_SOCKET, Id_SO_KEEPALIVE, iif(FKeepAlive, 1, 0));
```
rlebeau, line 1031: `TIdSocketHandle` has a public `SetKeepAliveValues()` method, you don't need to issue `SO_KEEPALIVE` directly here.

Lib/Protocols/IdHL7.pas
```diff
@@ -919,34 +1062,51 @@ procedure TIdHL7.StopServer;
...
+  sIp : String;
+  iIp : Cardinal;
 begin
   assert(Assigned(Self));
-  assert(Assigned(AContext));
+  Assert(Assigned(AThread));
+  //assert(Assigned(AContext));
   assert(Assigned(FLock));
+  LConnCount := 0;
+  sIp := (AThread.Connection.IOHandler as TIdIOHandlerSocket).Binding.PeerIP;
```
rlebeau, line 1078: Rather than type-casting the `IOHandler`, use the public `TIdTCPConnection.Socket` property instead.  Feel free to validate that `Socket` is not `nil`, since a non-socket `IOHandler` could be assigned for debugging purposes.

Lib/Protocols/IdHL7.pas
```diff
@@ -1067,26 +1243,34 @@ procedure TIdHL7.StartClient;
...
+		FClientThread.FClient.Disconnect;
+		FClientThread.FCloseEvent.SetEvent;
+      End
+    Else
+      InternalSetStatus(isStopped, 'Stopped');
   finally
     FLock.Leave;
     end;
-  LWaitStop := Ticks + 5000;
+  LStartTime := GetTickCount;
```
rlebeau, line 1265: Don't use `GetTickCount()` directly, it is not cross-platform.  Use `IdGlobal.Ticks()` instead.

Lib/Protocols/IdHL7.pas
```diff
@@ -1067,26 +1243,34 @@ procedure TIdHL7.StartClient;
...
+//  LWaitStop := Ticks + 5000;
   repeat
     LFinished := (GetStatus = IsStopped);
     if not LFinished then
       begin
       IndySleep(10);
       end;
-  until LFinished or (Ticks > LWaitStop);
+  Until LFinished Or (GetTickDiff(LStartTime,GetTickCount) > WAIT_STOP);
```
rlebeau, line 1273: `GetTickDiff()` requires use of `IdGlobal.Ticks()`, not `GetTickCount()`

Lib/Protocols/IdHL7.pas
```diff
@@ -1222,6 +1430,11 @@ procedure TIdHL7ClientThread.Execute;
           begin
           exit;
           end;
+        if FOwner.FKeepAlive Then
+        Begin
+          d := $FFFFFFFF;
+          FCLient.Socket.Binding.SetSockOpt(Id_SOL_SOCKET, Id_SO_KEEPALIVE, d);
```
rlebeau, line 1436: Use `TIdSocketHandle.SetKeepAliveValues()` instead.

Lib/Protocols/IdHL7.pas
```diff
@@ -1249,12 +1465,18 @@ procedure TIdHL7ClientThread.Execute;
...
           end;
-        if not Terminated then
+        If TimedOut Then
+          Begin
+          FClient.Disconnect;
+          End
+        Else If Not Terminated Then
           begin
           // we got disconnected. ReconnectDelay applies.
+          LRecTime := Now + ((FOwner.FReconnectDelay / 1000) * {second length} (1 / (24 * 60 * 60)));
```
rlebeau, line 1475: Prefer to use `DateUtils` functions, when available, to manipulate `TDateTime` values

Lib/Protocols/IdHL7.pas
```diff
@@ -1582,24 +1843,35 @@ function TIdHL7.GetMessage(var VMsg: String): TObject;
...
     end;
-  qm.FEvent.SetEvent;
+ //qm.FEvent.SetEvent;
 end;
 
+Function TIdHL7ClientThread.TimedOut: Boolean;
+Var
+  lGap : TDateTime;
+Begin
+  lGap := (now - FLastTraffic) * 24 * 60 * 60 * 1000;
```
rlebeau, line 1872: Prefer to use `DateUtils` functions, when available, to manipulate `TDateTime` values

rlebeau (changes requested):
Lots of little changes needed before I'll consider merging this into the main code.

Lib/Protocols/IdHL7.pas
```diff
@@ -184,25 +186,30 @@   EHL7CommunicationError = class(EIdException)
...
   MILLISECOND_LENGTH = (1 / (24 * 60 * 60 * 1000));
+  SEND_RESPONSE_NAMES : Array [TSendResponse] Of String = ('None', 'Error', 'NoConnection', 'Sent', 'OK', 'Timeout');
 
 type
   // the connection is provided in these events so that applications can obtain information about the
   // the peer. It's never OK to write to these connections
-  TMessageArriveEvent = procedure(ASender: TObject; AConnection: TIdTCPConnection; AMsg: String) of object;
-  TMessageReceiveEvent = procedure(ASender: TObject; AConnection: TIdTCPConnection; AMsg: String; var VHandled: Boolean; var VReply: String) of object;
-  TReceiveErrorEvent = procedure(ASender: TObject; AConnection: TIdTCPConnection; AMsg: String; AException: Exception; var VReply: String; var VDropConnection: Boolean) of object;
+  TMessageArriveEvent = Procedure(ASender: TObject; AConnection: TIdTCPConnection; AMsg: AnsiString) Of Object;
```
rlebeau, line 198: I'm not a fan of forcing `AnsiString` everywhere.  The rest of Indy uses the native `String` type, converting to/from bytes using the `IIdTextEncoding` interface as needed.  Most Internet protocols are 7bit/8bit data, and Indy handles them just fine, even in Unicode environments.  I would prefer that this unit does the same, for consistency with the rest of the Indy library.  Also, because `AnsiString` is charset-aware in D2009+, but this code is not handling charsets at all.  If specific charsets are needed, consider introducing `IIdTextEncoding` properties/parameters as needed to let the caller decide which charsets to use when conversions are needed.

shoraitek: Made changes regarding this, introduced DefSendEncoding now to address this (tested works nice)
https://github.com/IndySockets/Indy/pull/442/commits/7bd8d29be4574608893391b97fcb794d959042b9

Lib/Protocols/IdHL7.pas
shoraitek: Made change regarding this as far as possible for the logic
https://github.com/IndySockets/Indy/pull/442/commits/7bd8d29be4574608893391b97fcb794d959042b9

Lib/Protocols/IdHL7.pas
shoraitek: changed as requested
https://github.com/IndySockets/Indy/pull/442/commits/7bd8d29be4574608893391b97fcb794d959042b9

Lib/Protocols/IdHL7.pas
shoraitek: Changed as requested
https://github.com/IndySockets/Indy/pull/442/commits/7bd8d29be4574608893391b97fcb794d959042b9

Lib/Protocols/IdHL7.pas
shoraitek: Changed as requested
https://github.com/IndySockets/Indy/pull/442/commits/7bd8d29be4574608893391b97fcb794d959042b9

Lib/Protocols/IdHL7.pas
shoraitek: Changed as requested
https://github.com/IndySockets/Indy/pull/442/commits/7bd8d29be4574608893391b97fcb794d959042b9

Lib/Protocols/IdHL7.pas
shoraitek: Changed as requested
https://github.com/IndySockets/Indy/pull/442/commits/7bd8d29be4574608893391b97fcb794d959042b9

Lib/Protocols/IdHL7.pas
shoraitek: Changed as requested
https://github.com/IndySockets/Indy/pull/442/commits/7bd8d29be4574608893391b97fcb794d959042b9

Lib/Protocols/IdHL7.pas
shoraitek: Changed as requested
https://github.com/IndySockets/Indy/pull/442/commits/7bd8d29be4574608893391b97fcb794d959042b9

Lib/Protocols/IdHL7.pas
shoraitek: Changed as requested
https://github.com/IndySockets/Indy/pull/442/commits/7bd8d29be4574608893391b97fcb794d959042b9

Lib/Protocols/IdHL7.pas
shoraitek: Changed as requested
https://github.com/IndySockets/Indy/pull/442/commits/7bd8d29be4574608893391b97fcb794d959042b9

Lib/Protocols/IdHL7.pas
shoraitek: Changed as requested
https://github.com/IndySockets/Indy/pull/442/commits/7bd8d29be4574608893391b97fcb794d959042b9

Lib/Protocols/IdHL7.pas
shoraitek: Changed as requested
https://github.com/IndySockets/Indy/pull/442/commits/7bd8d29be4574608893391b97fcb794d959042b9

shoraitek:
Made load of changes as requested tested client and server to ensure utf8 character sending and receiving works nice
used 
https://github.com/RobHolme/HL7-Listener
to confirm the encoding received and test app to test if receiving.

Hopefully most issues resolved now, thank you for checking it all help and helping to improve @rlebeau

shoraitek:
Added example program to help with testing and usage in future
https://github.com/IndySockets/Indy/pull/442/commits/9442625acef973810fe7ac8218e921d55c32d5b3

Lib/Protocols/IdHL7.pas
shoraitek: Based on original authors code (new 2013 one), he seems to expect multiple message blocks in the buffer not just he one (Extracts the one block out of the buffer as needed, thus the Var ). Honestly we can look to change this logic maybe in future if there might be a reason for it as I cannot think of a scenario currently where he will get multiple blocks like this except if somehow he wants to allow bulk hl7 upload via one connection. In short it is as the original author had it and I kept it as is. hope it helps

rlebeau:
Thanks, I will review your changes shortly and let you know.

rlebeau:
I spent some time today reviewing your code, and making some changes to it.  I'll send you my revised version when it is ready.

shoraitek:
100% , thank you

rlebeau:
I pushed my changes into your branch, let me know what you think, and if I broke anything.

shoraitek:
> I pushed my changes into your branch, let me know what you think, and if I broke anything.

Hi there, I only use Lazarus to compile (currently on 2.2.4)

```
IdHL7.pas(507,22) Error: No matching implementation for interface method "_AddRef:LongInt; StdCall;" found
IdHL7.pas(507,22) Error: No matching implementation for interface method "_Release:LongInt; StdCall;" found
IdHL7.pas(562,3) Error: Identifier not found "Resut"
```

The not found Resut I fixed already, but the other to Addref and Release if I use TInterfacedObject directly then it compiles, but with TIdInterfacedObject I get the above error. Will look further into why this is happening

shoraitek:
I made changes to IdGlobalProtocols.pas for TIdInterfacedObject, please review if it is allowed please , made for fpc specific via defs

rlebeau:
Please don't put revision notes in the file headers, we don't do that anymore, and haven't done so for a very long time.  Revision notes belong in VCS checkins only.

That being said, I've changed `TIdQueuedMessage` back to using `TIdInterfacedObject` instead of `TInterfacedObject` directly.  That change will have to wait until Indy 11, when .NET support is dropped.

shoraitek:
> Please don't put revision notes in the file headers, we don't do that anymore, and haven't done so for a very long time. Revision notes belong in VCS checkins only.
> 
> That being said, I've changed `TIdQueuedMessage` back to using `TIdInterfacedObject` instead of `TInterfacedObject` directly. That change will have to wait until Indy 11, when .NET support is dropped.

Sorry about that. Will keep this in mind thank you

shoraitek:
@rlebeau  something odd now, will debug further this week. But basically if you use the sample app with these steps
A)Server side Click Listen
B) Client side click send
C) Server side click stop

Then the app will halt (after the connection close gracefully errors). debugging wise the server execute thread never seem to complete the Server Stop and the client is stuck trying to re-connect.

Is odd this happening, thought it might be due to server and client in same app and critical sections, but dont think it the case. I did test with running 2x programs (one server and one client), and it server side that hangs. Thus the server thread on disconnect is doing something odd. 

Thanks for all the help

shoraitek:
Some feedback I could trace the lock in the application it seems to TIdCustomTCPServer.SetActive -> TIdSchedulerOfThread.TerminateYarn  that just goes into a loop when called as per the callstack below. trying to figure out why

image

rlebeau:
When deactivating `TIdTCPServer`, it waits for its active threads to fully terminate themselves and remove their associated yarns from its `Scheduler`'s list.  Make sure that you are not deadlocking the server's threads, or doing anything that is thread-unsafe.

shoraitek:
> When deactivating `TIdTCPServer`, it waits for its active threads to fully terminate themselves and remove their associated yarns from its `Scheduler`'s list. Make sure that you are not deadlocking the server's threads, or doing anything that is thread-unsafe.

I changed example app to work but I added comments to the
TIdHL7.ServerDisconnect
function regarding the two events Connection Count Change and Disconnect for the server. Basically no matter what I do client side to try and make the write to the memobox threadsafe. I looked at TidNotify as an possible option but current have to rethink how to implement it from the component to the calling app.

shoraitek:
added tidnotify logic 
https://github.com/IndySockets/Indy/pull/442/commits/9eece92bbec9b85113158940ca38215ba21016a5
based on your answer in a question here
https://stackoverflow.com/questions/17506615/how-to-handle-log-in-a-threaded-manner-in-delphi

works well now thank you

rlebeau:
> added tidnotify logic [9eece92](https://github.com/IndySockets/Indy/commit/9eece92bbec9b85113158940ca38215ba21016a5) based on your answer in a question here https://stackoverflow.com/questions/17506615/how-to-handle-log-in-a-threaded-manner-in-delphi

See the update I just posted to [that answer](https://stackoverflow.com/a/17509002/65863).

> works well now thank you

Does that mean everything is now working in `TIdHL7` and it is ready to merge?

rlebeau:
In [b414507](https://github.com/IndySockets/Indy/pull/442/commits/b4145070a0b045e7bf707578360cf3111fa097c3#diff-a31d3755fad3ba2f2cd5060e618e597f2e5d5073d285aa0d5384e41f043d84c6), why did you feel the need to lowercase all of the native data types?

shoraitek:
Happend when I used the ctrl+d to format the code (Jedi format) while
trying to figure the issue out, dit not believe it would be to big a issue
and left it as is at that point

shoraitek:
I believe so yes, thank you for all the help and guidance

shoraitek:
> In [b414507](https://github.com/IndySockets/Indy/pull/442/commits/b4145070a0b045e7bf707578360cf3111fa097c3#diff-a31d3755fad3ba2f2cd5060e618e597f2e5d5073d285aa0d5384e41f043d84c6), why did you feel the need to lowercase all of the native data types used throughout the entire `IdHL7.pas` unit? Pascal is case-insensitive, that was a huge change for absolutely zero benefit. I'm tempted to just revert that change.

(sorry previous replies was from email). This is my settings currently, will see if I can get it to correct it again.
image

shoraitek:
Will this be more acceptable ? If so, I will book it in
image

image

rlebeau (approved):

Issue #445, opened 2023-01-27
Soket limit for Select on Linux (memory corruption)


MaxRusov:
TIdSocketListVCLPosix.Add checks Count >= FD_SETSIZE. It is incorrect in Posix. Need to check: AHandle < FD_SETSIZE.

_FD_SET causes memory corruption if AHandle > 1024. 

It is advisable to replace `select` to `poll` in Linux

rlebeau:
Is this working now?

MaxRusov:
This is not enough for my purposes (WEB Server). I have a lot more than 1024 connections.

In my version, I use `pool`, it seems to work.  But I'm handling a special case - only one socket at a time (Count = 1)

rlebeau:
What kind of code are you using that you are putting 1024 sockets into Indy's `TIdSocketList...`?  All of Indy's own internal use of `TIdSocketList...` only uses 1-2 sockets at a time per list instance.

rlebeau:
Internally, Indy uses only 1-2 sockets per `TIdSocketList...` instance.  But I get your point.  Indy will have to be updated in a future version to use `poll()` instead.

dbcto:
Does the problem also exist in the Indy version that comes with Delphi 11? We also use Indy as a web server under Linux.

rlebeau:
> Does the problem also exist in the Indy version that comes with Delphi 11?

Yes, it does.

MaxRusov:
> Per machine? Then how are you not running into file descriptor limitations outside of `select()`?
Per process. Limit for file descriptors can be increased by 'ulimit'

Note. Each outgoing connection also occupies one handle. If process made 1024 outgoing connection - first incoming connection got Handle=1025 and can't be 'select'ed. OOPS. My server made some outgoing connection for each incoming, so limit much more then 1024 'users'

dbcto:
> > Does the problem also exist in the Indy version that comes with Delphi 11?
> 
> Yes, it does. And Embarcadero is already beta testing the next RAD Studio release, so a fix for this issue will likely not make it into that release, either. Maybe the release after that...

That's good news. However, a new release of Embarcadero may take quite a while before it is released. 
Is there a relatively simple way or workaround to patch this in the current version?

rlebeau:
> Is there a relatively simple way or workaround to patch this in the current version?

[...]

PR #473, opened 2023-04-26
Change TIdSocketListVCLPosix to use poll instead of select

msccto:
fix for #445 

I replaced TIdSocketListVCLPosix with the code you suggested in the #445


Lib/System/IdStackVCLPosix.pas
zTorden, line 1: nfds is defined as nfds_t, which is unsigned long int, and that's UInt64, not integer

rlebeau: Technically, `nfds_t` is simply defined as an "unsigned integer".  Some platforms do use `unsigned long`, but there are others that use `unsigned int` instead.

zTorden: Just as i said t's defined in glibc as 'typedef unsigned long int nfds_t', but i do agree that there may be other definitions on different platforms, so i'd suggest using nativeuint here instead of integer

rlebeau: `NativeUInt` probably would not be appropriate either, because of the platforms that use `unsigned int` which doesn't change size between 32bit/64bit builds.  There really needs to be separate platform-appropriate `nfds_t` definitions.

zTorden: you're right, and this is actually the case with glibc where it is always unsigned long int both on 32 and 64 bit. still maybe unsigned long int itself differs on 32bit, can't really tell. but on ubuntu 22.04 which is always 64bit it's UInt64, i've checked that. but who knows, maybe it's only valid for gcc and not some other compiler, or is dependent on some 'compatibility' compiler options. Delphi defines this parameter as integer in Posix.SysSelect

zTorden: as for the de-facto standard System V AMD64 calling convention implemented on most (if not all) *nixes it's safe to be defined as integer (at least when there are less then 2^32 fds allocated), because even when it is passed via EDI register, the higher bits of RDI are being zeroed rendering it a valid 64bit value. e.g. after these operations
```
mov rdi, $ffffffffffffffff
mov edi, 0
```
the value of rdi would be 0 and not $ffffffff00000000

zTorden: but this is only valid for register-passed parameters (i.e. first 6 function parameters), if it's passed via stack (7+ parameter), or as a part of structure, or referenced by pointer it will cause trouble

rlebeau: In general, yes.  But that should not be a problem in THIS case since `poll()` has only 3 parameters and they are all simply types (1 pointer and 2 integers).  The 1st parameter is a pointer to a struct, but that does not affect the `nfds_t` parameter in question.

Issue #208, opened 2018-04-18
Issue with SASL use of Initial-Response parameter


rlebeau:
`TIdDICT`, `TIdIMAP4`, `TIdPOP3` and `TIdSMTP` implement SASL authentication.

It is currently unknown how the following issue affects `TIdDICT`, if at all.

`TIdIMAP4` uses SASL's Initial-Response only if the server advertises the IMAP `SASL-IR` capability.  No issue there.

POP3 and SMTP are affected by the same issue.  POP3 and SMTP servers have no way to advertise their support for Initial-Response.  `TIdPOP3` gets around this by simply disabling use of Initial-Response altogether.  But `TIdSMTP` currently enables Initial-Response unconditionally, which causes a problem.

When `TIdSMTP.AuthType` is set to `satSASL` (the issue does not affect `satDefault`), and `TIdSASLLogin` is enabled in the `TIdSMTP.SASLMechanisms` collection, `TIdSMTP` sends an `AUTH LOGIN` command where the username is passed in Initial-Response, like this:

```AUTH LOGIN usernameInBase64```

Not all SMTP servers support Initial-Response, though.  If such a server replies with a username prompt instead of a password prompt, eg:

```334 VXNlcm5hbWU6``` (`"Username:"` in base64)

`TIdSMTP` then sends the password instead of the username, and then the server replies with a password prompt (if not a failure due to an unknown username):

```334 UGFzc3dvcmQ6``` (`"Password:"` in base64)

`TIdSMTP` then sends the password again, and authentication usually fails.

Many 3rd party SMTP clients do not use Initial-Response, mainly because there is currently no defined way for an SMTP server to advertise its support for Initial-Response.  However, Initial-Response is formally defined as part of RFCs [2554](https://tools.ietf.org/html/rfc2554) and [4954](https://tools.ietf.org/html/rfc4954), "SMTP Service Extension for Authentication", and many SMTP servers do support Initial-Response.

Indy SASL components that currently implement Initial-Response, and thus are potentially affected by this issue, are:

```
TIdSASLAnonymous
TIdSASLExternal
TIdSASLLogin
TIdSASLNTLM
TIdSASLOTP
TIdSASLPlain
TIdSASLSKey
```

jaenicke:
I added pull request #530 to handle this bug.

jaenicke:
[...]

PR #530, opened 2024-04-11
Fix for #208 - TIdSMTP - correctly handle answer, if Initial-Response is not supported

jaenicke:
Fix for #208 - If Initial-Response is not supported, the server might fail to handle the request completely or the server might ask for the username, because the server does not handle the provided username after AUTH LOGIN.

At the moment, the login is cancelled at this moment, so it does not work at all.

So I added functionality to answer the request for the username correctly and continue.
**Note:**
The encoded string 'Username:' is hardcoded directly in the code at the moment. A constant would be better, but I did not know where to put it correctly.

In addition I added a property CanAttemptIR to TIdSMTP, so one can switch off the try with IR in case the server cannot handle this at all. This was supported already, but the functionality was not exposed, because this parameter was always set to True.

### DIFF
--- a/Lib/Protocols/IdSASLCollection.pas
+++ b/Lib/Protocols/IdSASLCollection.pas
@@ -229,6 +229,15 @@ begin
           Exit; // this mechanism is not supported
         end;
       end else begin
+        if TrimRight(AClient.LastCmdResult.Text.Text) = 'VXNlcm5hbWU6' then // 'Username:'
+        begin
+          // request for username --> Initial-Response failed, we have to send username again
+          AClient.SendCmd(AEncoder.Encode(S), []);//[334, 504]);
+          if CheckStrFail(AClient.LastCmdResult.Code, AOkReplies, AContinueReplies) then begin
+            ASASL.FinishAuthenticate;
+            Exit; // this mechanism is not supported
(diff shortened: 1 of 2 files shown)


rlebeau:
The change you propose for `IdSASLCollection.pas` really belongs in `IdSASLLogin.pas` instead, as such prompt handling is specific to the `AUTH LOGIN` command and not to other SASLs.

However, better handling of the `AUTH LOGIN` prompts was implemented in `IdSASLLogin.pas` 2 years ago in commit [c9cf49](https://github.com/IndySockets/Indy/commit/c9cf49ad464168fbf1a422c917aaec32b186e716), so this ticket really no longer applies to `TIdSASLLogin` and is still open as other SASLs still need to be reviewed for IR handling.

In the existing [`sasl-oauth`](https://github.com/IndySockets/Indy/tree/sasl-oauth) branch, there is a new `SASLCanAttemptInitialResponse` property added to `TIdPOP3`.  Such a property could be added to `TIdSMTP`, but I don't think it is be necessary since SMTP doesn't have the same problem with detecting IR support that POP3 has.  `IdSASLCollection.pas` already has logic in place to handle IR failure in POP3.

PR #542, opened 2024-08-10
Update IdCompressorZLib.pas

Ye-HO:
1. Z_BUF_ERROR is just a warning, not an error, it should be ignored.
2. Since a newer version of zlib is used, LZstream needs to be changed from passing value ​​to passing address.

### DIFF
--- a/Lib/Protocols/IdCompressorZLib.pas
+++ b/Lib/Protocols/IdCompressorZLib.pas
@@ -70,7 +70,7 @@ type
   TIdCompressorZLib = class(TIdZLibCompressorBase)
   protected
     function GetIsReady : Boolean; override;
-    procedure InternalDecompressStream(LZstream: TZStreamRec; AIOHandler : TIdIOHandler;
+    procedure InternalDecompressStream(var LZstream: TZStreamRec; AIOHandler : TIdIOHandler;
       AOutStream: TStream);
   public
 
@@ -105,7 +105,7 @@ const
 { TIdCompressorZLib }
 
 procedure TIdCompressorZLib.InternalDecompressStream(
-  LZstream: TZStreamRec; AIOHandler: TIdIOHandler; AOutStream: TStream);
+  var LZstream: TZStreamRec; AIOHandler: TIdIOHandler; AOutStream: TStream);
 {Note that much of this is taken from the ZLibEx unit and adapted to use the IOHandler}
 var
   zresult  : Integer;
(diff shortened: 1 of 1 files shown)


Lib/Protocols/IdCompressorZLib.pas
```diff
@@ -172,7 +172,11 @@ procedure TIdCompressorZLib.InternalDecompressStream(
     repeat
       LZstream.next_out := @outBuffer[0];
       LZstream.avail_out := bufferSize;
-      DCheck(inflate(LZstream,Z_NO_FLUSH));
+      zresult := inflate(LZstream, Z_NO_FLUSH);
+      if zresult <> Z_BUF_ERROR then
```
rlebeau, line 176: I understand that `Z_BUF_ERROR` is not a fatal error from `inflate`'s POV (there is a big comment in the code stating as much), but in this context isn't it really a fatal error? It is being given a 32K output buffer.  If "no progress was possible" at all with that much space provided to it, then how is ignoring `Z_BUF_ERROR` supposed to allow it to make progress?  Are you suggesting that it is still progressing through the input buffer, just not writing anything at all to the output buffer?

Lib/Protocols/IdCompressorZLib.pas
```diff
@@ -172,7 +172,11 @@ procedure TIdCompressorZLib.InternalDecompressStream(
...
       LZstream.next_out := @outBuffer[0];
       LZstream.avail_out := bufferSize;
-      DCheck(inflate(LZstream,Z_NO_FLUSH));
+      zresult := inflate(LZstream, Z_NO_FLUSH);
+      if zresult <> Z_BUF_ERROR then
+      begin
+        DCheck(zresult);
+      end;
       outSize := bufferSize - LZstream.avail_out;
       AOutStream.Write(outBuffer, outSize);
```
rlebeau, line 181: If `Z_BUF_ERROR` were to be ignored, then presumably `outSize` will be 0 in that situation, so moving `AOutStream.Write()` inside the `if` block would make sense, yes?

Lib/Protocols/IdCompressorZLib.pas
```diff
@@ -172,7 +172,11 @@ procedure TIdCompressorZLib.InternalDecompressStream(
...
     until (LZstream.avail_in = 0) and (LZstream.avail_out > 0);
```
rlebeau, line 182: Is there ever a situation where ignoring `Z_BUF_ERROR` may lead to an endless loop here?

Ye-HO: This has never happened in my work

Lib/Protocols/IdCompressorZLib.pas
Ye-HO: Delphi do what I did (System.ZLib.pas):
function ZDecompressCheckWithoutBufferError(code: Integer): Integer; overload;
begin
  Result := code;

  if code < 0 then
    if (code <> Z_BUF_ERROR) then
     raise EZDecompressionError.Create(string(_z_errmsg[2 - code])) at ReturnAddress;
end;

Lib/Protocols/IdCompressorZLib.pas
Ye-HO: I think thers's no need to do

Ye-HO: I think thers's no need to do so

rlebeau (approved):

PR #550, opened 2024-10-31
TIdFTP and TIdFTPServer - Implement CSID command

JPeterMugaas:
Specified by https://www.ietf.org/archive/id/draft-peterson-streamlined-ftp-command-extensions-10.txt and supported by WinSCP and CompleteFTP Server.

### DIFF
--- a/Lib/Protocols/IdFTP.pas
+++ b/Lib/Protocols/IdFTP.pas
@@ -1,4 +1,4 @@
-{
+﻿{
   $Project$
   $Workfile$
   $Revision$
@@ -674,52 +674,6 @@ type
   TIdNeedAccountEvent = procedure(ASender: TObject; var VAcct: string) of object;
 
   TIdFTPBannerEvent = procedure (ASender: TObject; const AMsg : String) of object;
-
-  TIdFTPClientIdentifier = class (TPersistent)
-  protected
-    FClientName : String;
-    FClientVersion : String;
-    FPlatformDescription : String;
-    FVendor : String;
-    function GetCSIDOutput: String;
-    procedure SetClientName(const AValue: String);
-    procedure SetClientVersion(const AValue: String);
-    procedure SetPlatformDescription(const AValue: String);
-    procedure SetVendor(const AValue : String);
-    function GetClntOutput: String;
-  public
-    procedure Assign(Source: TPersistent); override;
-    property ClntOutput : String read GetClntOutput;
-    property CSIDOutput : String read GetCSIDOutput;
-  published
-    property ClientName : String read FClientName write SetClientName;
-    property ClientVersion : String read FClientVersion write SetClientVersion;
-    property PlatformDescription : String read FPlatformDescription
-      write SetPlatformDescription;
-    property Vendor : String read FVendor write SetVendor;
-  end;
-  TIdFTPServerIdentifier = class(TObject)
-  protected
-    FServerInfo : TStrings;
-    function GetCaseSensitive: Boolean;
-    function GetPlatformVersion: String;
-    function GetPlatformDescription: String;
-    function GetServerName: String;
-    function GetServerVersion: String;
-    function GetVendor: String;
-    function GetDirSeparator : String;
-  public
-    constructor Create;
-    destructor Destroy; override;
-    property ServerInfo : TStrings read FServerInfo;
-    property ServerName : String read GetServerName;
-    property ServerVersion : String read GetServerVersion;
-    property Vendor : String read GetVendor;
-    property PlatformDescription : String read GetPlatformDescription;
-    property PlatformVersion : String read GetPlatformVersion;
-    property CaseSensitive : Boolean read GetCaseSensitive;
-    property DirSeparator : String read GetDirSeparator;
-  end;
 
   TIdFtpProxySettings = class (TPersistent)
   protected
@@ -2712,8 +2666,8 @@ end;
 
 destructor TIdFTP.Destroy;
 begin
-  FreeAndNil(FServerInfo);
   FreeAndNil(FClientInfo);
+  FreeAndNil(FServerInfo);
   FreeAndNil(FListResult);
   FreeAndNil(FLoginMsg);
   FreeAndNil(FDirectoryListing);
@@ -2771,22 +2725,20 @@ begin
       Break;
     end;
   end;
-  FServerInfo.ServerInfo.Clear;
-  if IsExtSupported('CSID') then begin {do not localize}
-      SendCmd('CSID ' + FClientInfo.GetCSIDOutput);  {do not localize}
-      if Self.LastCmdResult.NumericCode = 200 then begin
-        SplitDelimitedString(LastCmdResult.Text.Text, FServerInfo.ServerInfo, True, ';')
-      end;
-  end
-  else
-  begin
-  // send the CLNT command before sending the OPTS UTF8 command.
+
+  // identify the client before sending the OPTS UTF8 command.
   // some servers need this in order to work around a bug in
   // Microsoft Internet Explorer's UTF-8 handling
-    if IsExtSupported('CLNT') then begin {do not localize}
-      SendCmd('CLNT ' + FClientInfo.ClntOutput);  {do not localize}
+  FServerInfo.Clear;
+  if IsExtSupported('CSID') then begin {do not localize}
+    if SendCmd('CSID ' + FClientInfo.CSIDParams) = 200 then begin {do not localize}
+      FServerInfo.CSIDParams := LastCmdResult.Text.Text;
     end;
+  end
+  else if IsExtSupported('CLNT') then begin {do not localize}
+    SendCmd('CLNT ' + FClientInfo.CLNTParams);  {do not localize}
   end;
+
   // RLebeau 4/26/2019: per RFC 2640, if the server reports the 'UTF8'
   // capability, it is REQUIRED to detect and accept UTF-8 encoded
   // paths/filenames in commands.  But, it is not REQUIRED to send UTF-8
@@ -3944,79 +3896,6 @@ begin
   end;
 end;
 
-{ TIdFTPClientIdentifier }
-
-procedure TIdFTPClientIdentifier.Assign(Source: TPersistent);
-var
-  LSource: TIdFTPClientIdentifier;
-begin
-  if Source is TIdFTPClientIdentifier then begin
-    LSource := TIdFTPClientIdentifier(Source);
-    ClientName  := LSource.ClientName;
-    ClientVersion := LSource.ClientVersion;
-    PlatformDescription := LSource.PlatformDescription;
-    Vendor := LSource.Vendor;
-  end else begin
-    inherited Assign(Source);
-  end;
-end;
-
-//assume syntax such as this:
-//214 Syntax: CLNT <sp> <client-name> <sp> <client-version> [<sp> <optional platform info>] (Set client name)
-function TIdFTPClientIdentifier.GetClntOutput: String;
-begin
-  if FClientName <> '' then begin
-    Result := FClientName;
-    if FClientVersion <> '' then begin
-      Result := Result + ' ' + FClientVersion;
-      if FPlatformDescription <> '' then begin
-        Result := Result + ' ' + FPlatformDescription;
-      end;
-    end;
-  end
-  else
-  begin
-     Result := gsIdProductName + ' ' + gsIdVersion;
-  end;
-end;
-
-function TIdFTPClientIdentifier.GetCSIDOutput: String;
-begin
-  if FClientName <> '' then begin
-    Result := 'Name='+FClientName;
-    if FClientVersion <> '' then begin
-      Result := Result + '; Version='+FClientVersion;
-    end;
-    if FVendor <> '' then begin
-      Result := Result + '; Vendor='+FVendor;
-    end;
-  end
-  else
-  begin
-    Result := 'Name='+gsIdProductName + '; Version=' + gsIdVersion;
-  end;
-end;
-
-procedure TIdFTPClientIdentifier.SetClientName(const AValue: String);
-begin
-  FClientName := Trim(AValue);
-  // Don't call Fetch;  it prevents multi-word client names
-end;
-
-procedure TIdFTPClientIdentifier.SetClientVersion(const AValue: String);
-begin
-  FClientVersion := Trim(AValue);
-end;
-
-procedure TIdFTPClientIdentifier.SetPlatformDescription(const AValue: String);
-begin
-  FPlatformDescription := Trim(AValue);
-end;
-
-procedure TIdFTPClientIdentifier.SetVendor(const AValue: String);
-begin
-  FVendor := Trim(AValue);
-end;
 
 {Note about SetTime procedures:
 
@@ -4564,61 +4443,4 @@ begin
   end;
 end;
 
-{ TIdFTPServerIdentifier }
-
-constructor TIdFTPServerIdentifier.Create;
-begin
-  inherited Create;
-  FServerInfo := TStringList.Create;
-{$IFDEF HAS_TStringList_CaseSensitive}
-  TStringList(FServerInfo).CaseSensitive := False;
-{$ENDIF}
-end;
-
-destructor TIdFTPServerIdentifier.Destroy;
-begin
-  FreeAndNil(FServerInfo);
-  inherited;
-end;
-
-function TIdFTPServerIdentifier.GetCaseSensitive: Boolean;
-begin
-  Result := FServerInfo.Values['CaseSensitive'] <> '0';
-end;
-
-function TIdFTPServerIdentifier.GetDirSeparator: String;
-begin
-  Result := FServerInfo.Values['DirSep'];
-  if Result = '' then begin
-    Result := '/';
-  end;
-end;
-
-function TIdFTPServerIdentifier.GetPlatformDescription: String;
-begin
-  Result := FServerInfo.Values['OS'];
-end;
-
-function TIdFTPServerIdentifier.GetPlatformVersion: String;
-begin
-  Result := FServerInfo.Values['OSVer'];
-end;
-
-function TIdFTPServerIdentifier.GetServerName: String;
-begin
-  Result := FServerInfo.Values['Name'];
-end;
-
-function TIdFTPServerIdentifier.GetServerVersion: String;
-begin
-  Result := FServerInfo.Values['Version'];
-end;
-
-function TIdFTPServerIdentifier.GetVendor: String;
-begin
-  Result := FServerInfo.Values['Vendor'];
-end;
-
 end.
-
-
--- a/Lib/Protocols/IdFTP.pas
+++ b/Lib/Protocols/IdFTP.pas
@@ -2779,6 +2779,7 @@ begin
       SendCmd('CSID ' + LClnt);  {do not localize}
       if Self.LastCmdResult.NumericCode = 200 then begin
         LBuf := TrimRight(Self.LastCmdResult.Text.Text);
+        FServerInfo.ServerInfo.Clear;
         repeat
            FServerInfo.ServerInfo.Add(TrimLeft(Fetch(LBuf,';')));
         until LBuf = '';
--- a/Lib/Protocols/IdFTP.pas
+++ b/Lib/Protocols/IdFTP.pas
@@ -3969,7 +3969,6 @@ end;
 //214 Syntax: CLNT <sp> <client-name> <sp> <client-version> [<sp> <optional platform info>] (Set client name)
 function TIdFTPClientIdentifier.GetClntOutput: String;
 begin
-  Result := gsIdProductName + ' ' + gsIdVersion;
   if FClientName <> '' then begin
     Result := FClientName;
     if FClientVersion <> '' then begin
@@ -3978,12 +3977,15 @@ begin
         Result := Result + ' ' + FPlatformDescription;
       end;
     end;
+  end
+  else
+  begin
+     Result := gsIdProductName + ' ' + gsIdVersion;
   end;
 end;
 
 function TIdFTPClientIdentifier.GetCSIDOutput: String;
 begin
-  Result := 'Name='+gsIdProductName + '; Version=' + gsIdVersion;
   if FClientName <> '' then begin
     Result := 'Name='+FClientName;
     if FClientVersion <> '' then begin
@@ -3992,6 +3994,10 @@ begin
     if FVendor <> '' then begin
       Result := Result + '; Vendor='+FVendor;
     end;
+  end
+  else
+  begin
+    Result := 'Name='+gsIdProductName + '; Version=' + gsIdVersion;
   end;
 end;
 
--- a/Lib/Protocols/IdFTP.pas
+++ b/Lib/Protocols/IdFTP.pas
@@ -701,12 +701,12 @@ type
   protected
     FServerInfo : TStrings;
     function GetCaseSensitive: Boolean;
-    function GetPlatformVer: String;
+    function GetPlatformVersion: String;
     function GetPlatformDescription: String;
     function GetServerName: String;
     function GetServerVersion: String;
     function GetVendor: String;
-    function GetDirSeperator : String;
+    function GetDirSeparator : String;
   public
     constructor Create;
     destructor Destroy; override;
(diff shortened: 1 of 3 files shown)


Lib/Protocols/IdFTP.pas
```diff
@@ -2742,18 +2764,45 @@ procedure TIdFTP.IssueFEAT;
...
+    if IsExtSupported('CSID') then begin {do not localize}
+      LClnt := 'Name='+FClientInfo.FClientName;
+      if FClientInfo.FClientVersion <> '' then begin
+        LClnt := LClnt + '; Version='+FClientInfo.FClientVersion;
+      end;
+      if FClientInfo.FVendor <> '' then begin
+        LClnt := LClnt + '; Vendor='+FClientInfo.FVendor;
+      end;
+      SendCmd('CSID ' + LClnt);  {do not localize}
+      LBuf := TrimRight(Self.LastCmdResult.Text.Text);
```
rlebeau, line 2777: If the `CSID` command fails, the response should not be parsed.  It may be a non-empty string but will not be in a format that your parser is expecting.

Lib/Protocols/IdFTP.pas
```diff
@@ -2742,18 +2764,45 @@ procedure TIdFTP.IssueFEAT;
...
+      FServerInfo.CaseSensitive := True;
+      repeat
+         LClnt := Fetch(LBuf,';');
+         LValName := TrimLeft(Fetch(LClnt,'='));
+         case PosInStrArray(LValName,['Name','Version','Vendor','OS','OSVer','CaseSensitive']) of
```
rlebeau, line 2782: A `CSID` response can contain more name=value pairs than just the few you are looking for.  I would suggest providing a catch-all property to store any extra values. It may make more sense to just store all of the received values into a `TStringList` and then have a few property getters that pull from that list.

Lib/Protocols/IdFTPServer.pas
```diff
@@ -739,6 +739,20 @@   EIdFTPCannotBeNegative = class(EIdFTPServerException);
 
   TIdFTPServer = class;
 
+  TIdFTPServerInfo = class(TPersistent)
```
rlebeau, line 742: A `CSID` response can contain additional name=value pairs (`os-dependant-fact`, `local-fact`, etc) so I would suggest adding a way for the user to provide those values as well.

Lib/Protocols/IdFTPServer.pas
```diff
@@ -6018,6 +6053,56 @@ function TIdFTPServer.MLSFEATLine(const AFactMask: TIdMLSDAttrs;
...
+   LBuf := AParams;
+   repeat
+      LValueValue := Fetch(LBuf,';');
+      LValueName := TrimLeft(Fetch(LValueValue,'='));
+      case PosInStrArray(LValueName,['Name','Version']) of
+        0 : LCltName := LValueValue;
+        1 : LCltVersion := LValueValue;
+      end;
+   until LBuf = '';
+   Result := Trim(LCltName + ' ' + LCltVersion);
```
rlebeau, line 6073: I would suggest including the `OS` value to mimic the `PlatformDescription` portion of the `CLNT` command.

Lib/Protocols/IdFTP.pas
```diff
@@ -2742,18 +2764,45 @@ procedure TIdFTP.IssueFEAT;
       Break;
     end;
   end;
-
+  if FClientInfo.ClientName <> '' then begin
+    if IsExtSupported('CSID') then begin {do not localize}
+      LClnt := 'Name='+FClientInfo.FClientName;
```
rlebeau, line 2769: Rather than having this logic directly in `IssueFEAT`, how about adding a `CsidOutput` property to `TIdFTPClientIdentifier` instead?

Lib/Protocols/IdFTPServer.pas
```diff
@@ -6018,6 +6053,56 @@ function TIdFTPServer.MLSFEATLine(const AFactMask: TIdMLSDAttrs;
...
+end;
+
+procedure TIdFTPServer.CommandCSID(ASender: TIdCommand);
```
rlebeau, line 6076: A CSID response can contain additional name=value pairs (os-dependant-fact, local-fact, etc) so I would suggest adding a way for the user to provide those values as well.

Lib/Protocols/IdFTPServer.pas
```diff
@@ -6059,6 +6144,23 @@ procedure TIdFTPServer.DoOnDataPortBeforeBind(ASender: TIdFTPServerContext);
   end;
 end;
 
+function TIdFTPServer.FTPIsCaseSensitive: Boolean;
+begin
+  case FPathProcessing of
+    ftppDOS : Result := False;
+    ftpOSDependent :
+      begin
+        if GOSType = otWindows then begin
```
rlebeau, line 6153: I would get rid of the `if` here: `Result := (GOSType <> otWindows);`

rlebeau (changes requested):
See inline comments

Lib/Protocols/IdFTPServer.pas
JPeterMugaas: if you mean, rename the variable OS, I have made that change.

Lib/Protocols/IdFTP.pas
```diff
@@ -2742,18 +2771,31 @@ procedure TIdFTP.IssueFEAT;
...
   // some servers need this in order to work around a bug in
   // Microsoft Internet Explorer's UTF-8 handling
-  if IsExtSupported('CLNT') then begin {do not localize}
-    LClnt := FClientInfo.ClntOutput;
-    if LClnt = '' then begin
-      LClnt := gsIdProductName + ' ' + gsIdVersion;
+      if IsExtSupported('CLNT') then begin {do not localize}
+        LClnt := FClientInfo.ClntOutput;
+        if LClnt = '' then begin
+          LClnt := gsIdProductName + ' ' + gsIdVersion;
```
rlebeau, line 2793: Since you are now checking `FClientInfo.ClientName` beforehand, `FClientInfo.ClntOutput` can never return a blank string here anymore, so using Indy's product/version here is now dead code.

On the other hand, the original code would always send `CLNT` if supported by the server, and if `FClientInfo.ClientName` is empty then Indy's default product/version is sent.  That logic should probably be preserved.  In which case, I would suggest also making `CSID` use Indy's product/version as well if `FClientInfo.ClientName` is empty.

Lib/Protocols/IdFTPServer.pas
rlebeau: No, not rename. I meant if the client sends `OS`/`OSVer` in the `CSID` command then include that info in the string that you are passing to the `OnClientID` event, to mimic what `CommandCLNT` passes to `OnClientID` if that info  is present.

JPeterMugaas: I don't think there is a PlatformDescription that is sent to the server with the CLNT command.  FileZilla sends "CLNT FileZilla" and WS-FTP-Professional sends "CLNT WS_FTP_Professional 12.9.1".  And further, the draft https://datatracker.ietf.org/doc/html/draft-peterson-streamlined-ftp-comm  and-extensions-10 does not state that the client can send an "OS" as a parameter to the CSID command.

Lib/Protocols/IdFTP.pas
```diff
@@ -3988,7 +3989,10 @@ function TIdFTPClientIdentifier.GetClntOutput: String;
 
 function TIdFTPClientIdentifier.GetCSIDOutput: String;
 begin
-  Result := 'Name='+FClientName;
+  Result := '';
+  if FClientName <> '' then begin
+    Result := 'Name='+FClientName;
+  end;
   if FClientVersion <> '' then begin
     Result := Result + '; Version='+FClientVersion;
```
rlebeau, line 3998: If `FClientName` is empty, but `FClentVersion` and/or `FVendor` are not empty, then `GetCSIDOutput()` will return invalid output.

Lib/Protocols/IdFTP.pas
```diff
@@ -2771,29 +2771,30 @@ procedure TIdFTP.IssueFEAT;
...
     end;
   end;
-  if FClientInfo.ClientName <> '' then begin
-    if IsExtSupported('CSID') then begin {do not localize}
-
-      SendCmd('CSID ' + FClientInfo.GetCSIDOutput);  {do not localize}
+  if IsExtSupported('CSID') then begin {do not localize}
+      LClnt := FClientInfo.GetCSIDOutput;
+      if LClnt = '' then begin
+        LClnt := 'Name='+gsIdProductName + '; Version=' + gsIdVersion;
```
rlebeau, line 2777: The more I think about it, use of `gsIdProductName` and `gsIdVersion` should probably be moved inside of `GetCSIDOutput()` as the default return value when `FClientName` is empty.

Lib/Protocols/IdFTP.pas
```diff
@@ -2771,29 +2771,30 @@ procedure TIdFTP.IssueFEAT;
...
-      if IsExtSupported('CLNT') then begin {do not localize}
-        LClnt := FClientInfo.ClntOutput;
-        if LClnt = '' then begin
-          LClnt := gsIdProductName + ' ' + gsIdVersion;
-        end;
-        SendCmd('CLNT ' + LClnt);  {do not localize}
+    if IsExtSupported('CLNT') then begin {do not localize}
+      LClnt := FClientInfo.ClntOutput;
+      if LClnt = '' then begin
+        LClnt := gsIdProductName + ' ' + gsIdVersion;
```
rlebeau, line 2796: The more I think about it, use of `gsIdProductName` and `gsIdVersion` should probably be moved inside of `GetClntOutput()` as the default return value when `FClientName` is empty.

Lib/Protocols/IdFTP.pas
```diff
@@ -3975,6 +3969,7 @@ procedure TIdFTPClientIdentifier.Assign(Source: TPersistent);
 //214 Syntax: CLNT <sp> <client-name> <sp> <client-version> [<sp> <optional platform info>] (Set client name)
 function TIdFTPClientIdentifier.GetClntOutput: String;
 begin
+  Result := gsIdProductName + ' ' + gsIdVersion;
```
rlebeau, line 3972: Why put the default here instead of in the `else` block?

Lib/Protocols/IdFTP.pas
```diff
@@ -3983,22 +3978,20 @@ function TIdFTPClientIdentifier.GetClntOutput: String;
...
     end;
-  end else begin
-    Result := '';
   end;
 end;
 
 function TIdFTPClientIdentifier.GetCSIDOutput: String;
 begin
-  Result := '';
+  Result := 'Name='+gsIdProductName + '; Version=' + gsIdVersion;
```
rlebeau, line 3986: Why put the default here instead of in an `else` block when `FClientName` is empty?

Lib/Protocols/IdFTP.pas
```diff
@@ -680,17 +680,44 @@   TIdFTPClientIdentifier = class (TPersistent)
...
+    function GetDirSeperator : String;
+  public
+    constructor Create;
+    destructor Destroy; override;
+    property ServerInfo : TStrings read FServerInfo;
+    property ServerName : String read GetServerName;
+    property ServerVersion : String read GetServerVersion;
+    property Vendor : String read GetVendor;
+    property PlatformDescription : String read GetPlatformDescription;
+    property PlatformVer : String read GetPlatformVer;
```
rlebeau, line 718: Should rename to `PlatformVersion`

Lib/Protocols/IdFTP.pas
```diff
@@ -2742,18 +2771,27 @@ procedure TIdFTP.IssueFEAT;
       Break;
     end;
   end;
-
+  if IsExtSupported('CSID') then begin {do not localize}
+      LClnt := FClientInfo.GetCSIDOutput;
+      SendCmd('CSID ' + LClnt);  {do not localize}
+      if Self.LastCmdResult.NumericCode = 200 then begin
```
rlebeau, line 2777: This can be simplified to `if SendCmd('CSID ' + FClientInfo.CSIDOutput') = 200 then ...`

Lib/Protocols/IdFTP.pas
```diff
@@ -2742,18 +2771,27 @@ procedure TIdFTP.IssueFEAT;
       Break;
     end;
   end;
-
+  if IsExtSupported('CSID') then begin {do not localize}
+      LClnt := FClientInfo.GetCSIDOutput;
+      SendCmd('CSID ' + LClnt);  {do not localize}
+      if Self.LastCmdResult.NumericCode = 200 then begin
+        LBuf := TrimRight(Self.LastCmdResult.Text.Text);
+        FServerInfo.ServerInfo.Clear;
```
rlebeau, line 2779: The `ServerInfo` should be cleared before sending the `CSID` command. Or at least also cleared if the command fails.

Lib/Protocols/IdFTP.pas
```diff
@@ -4506,6 +4563,58 @@ function TIdFTP.VerifyFile(ALocalFile: TStream; const ARemoteFile: String;
   end;
 end;
 
+{ TIdFTPServerIdentifier }
+
+constructor TIdFTPServerIdentifier.Create;
+begin
+  inherited Create;
+  FServerInfo := TStringList.Create;
```
rlebeau, line 4571: `CSID` fact names are case-insensitive, so add this:

```
{$IFDEF HAS_TStringList_CaseSensitive}
TStringList(FServerInfo).CaseSensitive := False;
{$ENDIF}
```

Lib/Protocols/IdFTPServer.pas
```diff
@@ -739,6 +739,26 @@   EIdFTPCannotBeNegative = class(EIdFTPServerException);
...
+  public
+    constructor Create;
+    destructor Destroy; override;
+  published
+    property ServerName : String read FServerName write FServerName;
+    property ServerVersion : String read FServerVersion write FServerVersion;
+    property ServerVendor : String read FServerVendor write FServerVendor;
+    property PlatformDescription : String read FPlatformDescription write FPlatformDescription;
+    property PlatformVersion : String read FPlatformVersion write FPlatformVersion;
+    property AdditionalFacts : TStrings read FAdditionalFacts write SetAdditionalFacts;
```
rlebeau, line 760: Should `CaseSensitive` and `DirSeparator` be included as properties here, to match `TIdFTPServerIdentifier` in `IdFTP.pas`?

Lib/Protocols/IdFTPServer.pas
```diff
@@ -1009,6 +1030,8 @@   TIdFTPServer = class(TIdExplicitTLSServer)
     function DoProcessPath(ASender : TIdFTPServerContext; const APath: TIdFTPFileName): TIdFTPFileName;
 
     function FTPNormalizePath(const APath: String) : String;
+    function GetPathSeperator : String;
```
rlebeau, line 1033: Should be renamed to `GetPathSeparator`

Lib/Protocols/IdFTPServer.pas
```diff
@@ -6018,6 +6060,77 @@ function TIdFTPServer.MLSFEATLine(const AFactMask: TIdMLSDAttrs;
...
+     Result := Result + ' ' + LCltOS;
+   end;
+end;
+
+procedure TIdFTPServer.CommandCSID(ASender: TIdCommand);
+var LClientInfo : String;
+   LServerInfo : String;
+   i : Integer;
+begin
+  if Assigned(FOnClientID) then begin
```
rlebeau, line 6095: Should a new event be added (ie `OnClientIDEx`?) to expose all of the client's facts?

Lib/Protocols/IdFTPServer.pas
```diff
@@ -6018,6 +6060,77 @@ function TIdFTPServer.MLSFEATLine(const AFactMask: TIdMLSDAttrs;
...
+    LClientInfo := ParseCSIDParams(ASender.UnparsedParams);
+    FOnClientID(ASender.Context as TIdFTPServerContext, LClientInfo);
+  end;
+  if FServerInfo.ServerName <> '' then begin
```
rlebeau, line 6099: If `ServerName` is empty but any of the other facts are not, then the reply will be malformed.  The separating `;` tokens need to be added dynamically based on whether `LServerInfo` is blank or not.

Lib/Protocols/IdFTPServer.pas
```diff
@@ -6018,6 +6060,77 @@ function TIdFTPServer.MLSFEATLine(const AFactMask: TIdMLSDAttrs;
...
+    LServerInfo := LServerInfo + '0';
+  end;
+  //https://solarwindscore.my.site.com/SuccessCenter/s/article/CSID-FTP-command?language=en_US
+  //states that the DirSep fact is required to be reported.
+  LServerInfo := LServerInfo + '; DirSep='+GetPathSeperator;
+  for i := 0 to FServerInfo.AdditionalFacts.Count -1 do
+  begin
+    LServerInfo := LServerInfo + '; '+TrimLeft(FServerInfo.AdditionalFacts[i]);
+  end;
+  if StartsWith(LServerInfo,'; ') then
```
rlebeau, line 6127: Should be using `TextStartsWith()` instead

Lib/Protocols/IdFTPServer.pas
```diff
@@ -6059,6 +6172,34 @@ procedure TIdFTPServer.DoOnDataPortBeforeBind(ASender: TIdFTPServerContext);
   end;
 end;
 
+function TIdFTPServer.GetPathSeperator : String;
+begin
+  Result := '/';
+  case FPathProcessing of
```
rlebeau, line 6178: What about `ftppCustom`? Do we need a way for the user to indicate a custom separator?

Lib/Protocols/IdFTPServer.pas
```diff
@@ -6059,6 +6172,34 @@ procedure TIdFTPServer.DoOnDataPortBeforeBind(ASender: TIdFTPServerContext);
...
+        begin
+          Result := '\';
+        end;
+      end;
+  end;
+end;
+
+function TIdFTPServer.FTPIsCaseSensitive: Boolean;
+begin
+  case FPathProcessing of
```
rlebeau, line 6192: What about `ftppCustom`? Do we need a way for the user to indicate whether they are case-sensitive or not?

Lib/Protocols/IdFTPServer.pas
```diff
@@ -7373,4 +7514,23 @@ procedure TIdFTPServerContext.ResetZLibSettings;
   FZLibStratagy := DEF_ZLIB_STRATAGY; // - default
 end;
 
+{ TIdFTPServerInfo }
+
+constructor TIdFTPServerInfo.Create;
+begin
+  inherited Create;
+  FAdditionalFacts := TStringList.Create;
```
rlebeau, line 7522: `CSID` fact names are case-insensitive, so add this:

```
{$IFDEF HAS_TStringList_CaseSensitive}
TStringList(FAdditionalFacts).CaseSensitive := False;
{$ENDIF}
```

Lib/Protocols/IdFTPServer.pas
JPeterMugaas: We do in that case and I made two properties for the FTP Server for this.

Lib/Protocols/IdFTPServer.pas
```diff
@@ -3948,6 +3995,9 @@ procedure TIdFTPServer.CommandFEAT(ASender: TIdCommand);
   if (UseTLS <> utNoTLSSupport) and (LContext.Binding.IPVersion = Id_IPv4) then begin
     ASender.Reply.Text.Add('CPSV');   {Do not translate}
   end;
+  if Assigned(FOnClientID) then begin
+    ASender.Reply.Text.Add('CSID');  {Do not localize}
```
rlebeau, line 3999: Since a `CSID` reply is sent back to the client regardless of whether the `OnClientID` event is assigned, this `FEAT` entry should not be conditional.

Lib/Protocols/IdFTPServer.pas
```diff
@@ -6018,6 +6068,77 @@ function TIdFTPServer.MLSFEATLine(const AFactMask: TIdMLSDAttrs;
...
+  LServerInfo := LServerInfo + '; DirSep='+GetPathSeparator;
+  for i := 0 to FServerInfo.AdditionalFacts.Count -1 do
+  begin
+    LServerInfo := LServerInfo + '; '+TrimLeft(FServerInfo.AdditionalFacts[i]);
+  end;
+  if TextStartsWith(LServerInfo,'; ') then
+  begin
+    IdDelete(LServerInfo,1,2);
+  end;
+  ASender.Reply.SetReply( 200, LServerInfo);
```
rlebeau, line 6139: Per the `CSID` spec:

> The CSID command MUST be available only while the client-FTP process is logged in ... If the client-FTP process is not logged in, the server-FTP process MUST return a 530 response code."

So `CommandCSID()` should check for that condition.

Lib/Protocols/IdFTPServer.pas
```diff
@@ -7575,4 +7584,108 @@ procedure TIdFTPServerInfo.SetAdditionalFacts(const AValue: TStrings);
...
+  FFacts.Values['Version'] := AValue;
+end;
+
+procedure TIdFTPClientInfo.SetCLNTParams(const AValue: String);
+var LBuf : String;
+begin
+  LBuf := TrimLeft(Avalue);
+  ClientName := Fetch(LBuf);
+  ClientVersion := Fetch(LBuf);
+  PlatformDescription := Fetch(LBuf);
```
rlebeau, line 7661: Since nothing follows the description, there is no need to use `Fetch()` on this line.

Lib/Protocols/IdFTPServer.pas
```diff
@@ -7575,4 +7584,108 @@ procedure TIdFTPServerInfo.SetAdditionalFacts(const AValue: TStrings);
...
+end;
+
+procedure TIdFTPClientInfo.SetPlatformDescription(const AValue: String);
+begin
+  FFacts.Values['OS'] := AValue;
+end;
+
+procedure TIdFTPClientInfo.SetPlatformVersion(const AValue: String);
+begin
+  FFacts.Values['OS'] := AValue;
```
rlebeau: This need to be `OSVer` instead

Lib/Protocols/IdFTP.pas
```diff
@@ -2742,18 +2770,25 @@ procedure TIdFTP.IssueFEAT;
...
     end;
   end;
-
+  FServerInfo.ServerInfo.Clear;
+  if IsExtSupported('CSID') then begin {do not localize}
+      SendCmd('CSID ' + FClientInfo.GetCSIDOutput);  {do not localize}
+      if Self.LastCmdResult.NumericCode = 200 then begin
+        LBuf := TrimRight(Self.LastCmdResult.Text.Text);
+        repeat
+           FServerInfo.ServerInfo.Add(TrimLeft(Fetch(LBuf,';')));
```
rlebeau, line 2779: Perhaps replace this whole loop with a single call to `IdGlobal.SplitDelimitedString()` instead?

`SplitDelimitedString(LastCmdResult.Text.Text, FServer.ServerInfo, True, ';');`

Lib/Protocols/IdFTPServer.pas
```diff
@@ -7373,4 +7562,145 @@ procedure TIdFTPServerContext.ResetZLibSettings;
...
+end;
+
+procedure TIdFTPClientInfo.SetCSIDParams(const AValue: String);
+var LBuf,
+    LValue : String;
+begin
+   FFacts.Clear;
+   LBuf := AValue;
+   repeat
+      LValue := TrimLeft(Fetch(LBuf,';'));
```
rlebeau, line 7686: Perhaps replace this whole loop with a single call to `IdGlobal.SplitDelimitedString()` instead?

`SplitDelimitedString(AValue, FFacts, True, ';');`

Lib/Protocols/IdFTP.pas
```diff
@@ -680,17 +680,44 @@   TIdFTPClientIdentifier = class (TPersistent)
     FClientName : String;
     FClientVersion : String;
     FPlatformDescription : String;
+    FVendor : String;
```
rlebeau, line 683: Copy of `FVendor` is missing in `TIdFTPClientIdentifier.Assign()`

Lib/Protocols/IdFTP.pas
```diff
@@ -680,17 +680,44 @@   TIdFTPClientIdentifier = class (TPersistent)
...
     procedure Assign(Source: TPersistent); override;
     property ClntOutput : String read GetClntOutput;
+    property CSIDOutput : String read GetCSIDOutput;
   published
     property ClientName : String read FClientName write SetClientName;
     property ClientVersion : String read FClientVersion write SetClientVersion;
-    property PlatformDescription : String read FPlatformDescription write SetPlatformDescription;
+    property Vendor : String read FVendor write FVendor;
+    property PlatformDescription : String read FPlatformDescription
+      write FPlatformDescription;
```
rlebeau, line 698: Why was `Vendor` added before `PlatformDescription`? You removed `SetPlatformDescription()` from the `PlatformDescription` property setter, but did not remove `SetPlatformDescription()` itself from the unit.  This seems like a needless change. Should `Vendor` and `PlatformDescription` have setters to mirror the `ClientName` and ClientVersion` properties, which trim their input values?

rlebeau (approved):

PR #558, opened 2024-11-20
Remove some unneeded units from uses clause.   FreePascal/Lazarus war…

JPeterMugaas:
…ns about this.


Lib/Protocols/IdRegister.pas
```diff
@@ -277,9 +272,7 @@ implementation
   IdSASLOTP,
   IdSASLPlain,
   IdSASLSKey,
-  IdSASLUserPass,
   IdSASL_CRAM_MD5,
-  IdSASL_CRAM_SHA1,
```
rlebeau, line 282: Removing `IdSASL_CRAM_SHA1` causes an error in Delphi, as `TIdSASLCRAMSHA1` is being registered in Delphi but not in FPC.  So either `IFDEF` this unit or fix the registration of `TIdSASLCRAMSHA1` for FPC.

Lib/Protocols/IdRegister.pas
```diff
@@ -323,7 +316,6 @@ implementation
   IdUserAccounts,
   IdUserPassProvider,
   IdVCard,
-  IdWebDAV,
```
rlebeau, line 326: Removing `IdWebDAV` causes an error in Delphi, as `TIdWebDAV` is being registered in Delphi but not in FPC. So either `IFDEF` this unit or fix the registration of `TIdWebDAV` for FPC.

rlebeau (changes requested):
See comments

Lib/Core/IdRegisterCore.pas
```diff
@@ -137,7 +137,6 @@ implementation
   IdGlobal,
   {$ENDIF}
 
-  IdBaseComponent,
```
rlebeau, line 140: `TIdBaseComponent` is being used by `TIdBaseComponentSelectionEditor` in Delphi, so this unit needs to be `IFDEF`'ed on `HAS_TSelectionEditor`.

rlebeau (approved):

PR #559, opened 2024-11-21
Move mode switch from the interface part of the unit to the area betw…

JPeterMugaas:
…een the "unit" and "interface" for good measure

### DIFF
--- a/Lib/Core/IdAboutVCL.pas
+++ b/Lib/Core/IdAboutVCL.pas
@@ -11,11 +11,6 @@ uses
 {$IFDEF WIDGET_VCL_LIKE}
   StdCtrls, Buttons, ExtCtrls, Graphics, Controls, ComCtrls, Forms,
 {$ENDIF}
-{$IFDEF HAS_UNIT_Types}
-  {$IFNDEF FPC}
-  Types,
-  {$ENDIF}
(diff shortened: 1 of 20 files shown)


JPeterMugaas:
I moved the IdCompilerDefines.inc to between the "unit" and "imterface" declarations because of this:

From the Free Pascal Conversion Guide:
https://wiki.lazarus.freepascal.org/Code_Conversion_Guide

Most Delphi units can be compiled by the Free Pascal compiler by adding:

{$IFDEF FPC}
  {$MODE DELPHI}
{$ENDIF}
right after the unit name.

rlebeau:
Where the `.inc` reference is located right now is fine, it has worked fine for years without problem. It doesn't need to be moved:

https://www.freepascal.org/docs-html/user/userse33.html

> The `MODE` directive should always be located before the `uses` clause of the unit interface or program `uses` clause, because setting the mode may result in the loading of an additional unit as the first unit to be loaded.

Moving the `.inc` references above the `interface` block doesn't offer any real benefit, only minor code cleanup. I'm already doing some other code cleanups in Indy 11, so I would defer such a big change until then.

But, removing unused unit references is good.  I'll incorporate those changes accordingly.

PR #561, opened 2024-11-27
'zlib1.dll' and 'zlibwapi.dll' are no longer available on zlib.net's website

JPeterMugaas:
For FreePascal for Win32 and Win64, we need to provide ZLib DLL's. These are created by:

1) Downloading the ZLib source-code.

and

2) The source-code has a CMakeLists.txt.  Use that in Visual Studio 20022's CMake build using both a Win32 and Win64 release prfile.


Lib/Protocols/IdZLibHeaders.pas
```diff
@@ -892,19 +892,18 @@ function inflateGetHeader; external;
...
   libzlib = 'zlib';
   {$ENDIF}
+  //Do not use IFDEF WINDOWS because that may catch WindowsCE.
   {$IFDEF WIN32}
-  //Note that this is the official ZLIB1 .DLL from the http://www.zlib.net/
+  //Note that this is not an official ZLIB .DLL.  It was created by running
+  //CMake in Visual Studio 2022.
 const
-  libzlib = 'zlib1.dll';
+  zlib.dllzlib1.dll';
```
rlebeau, line 900: What is this line trying to do? This is clearly malformed.

rlebeau (approved):

PR #564, opened 2024-12-05
Fix warning about inlining something in IdCookie.

JPeterMugaas:


rlebeau:
Why is this needed?  What is the actual warning? When I compile the latest Indy in Delphi 12, I don't get any warnings in `IdCookie.pas` at all.  Are you compiling with a different version?

JPeterMugaas:
The Hint I got was:

[DCC Hint] IdCookie.pas(420): H2443 Inline function 'TList.Remove' has not been expanded because unit 'System.Types' is not specified in USES list

The IDE version I'm using is:

Embarcadero® Delphi 12 Version 29.0.53982.0329 
Copyright © 2024 Embarcadero Technologies, Inc. All Rights Reserved.

It has patch 2.

I'm compiling with the DEBUG settings for Win32.  The program source-code is available at:

https://github.com/TurboPack/indy_extras/tree/main/TaurusTLS/Demos/TaurusHTTPServer

rlebeau:
I am also using 29.0.53982.0329 with Patch 2, and I do not see that hint when I compile Indy for Win32 debug.

Also, there is no call to `TList.Remove()` on line 420 of `IdCookie.pas`.  The nearest call is on line 415.  This makes me wonder if you are using the correct version of Indy.

In any case, I have made a small adjustment to your branch code and will merge it.

rlebeau (approved):

PR #572, opened 2024-12-15
Fix code reversion that was breaking TaurusTLS.

JPeterMugaas:
move the types and constants back.


rlebeau:
I have corrected the problem separately and closing this merge as no longer needed.

PR #573, opened 2024-12-16
Ftp algs command

JPeterMugaas:
Command specified by:
https://www.rfc-editor.org/rfc/rfc6384.html#section-11

FTP servers (as opposed
   to ALGs) MUST NOT perform any actions upon receiving the ALGS
   command.  However, FTP servers MUST still send a response.  If FTP
   servers recognize the ALGS command, the best course of action would
   be to return a 202 response:

      202 Command not implemented, superfluous at this site

### DIFF
--- a/Lib/Protocols/IdFTPServer.pas
+++ b/Lib/Protocols/IdFTPServer.pas
@@ -2232,7 +2232,7 @@ begin
 
   //ALGS  - RFC 6384
   LCmd := CommandHandlers.Add;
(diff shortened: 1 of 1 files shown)


rlebeau:
Is this really needed? That same RFC says:

> However, there is no reason for FTP servers to specifically recognize this command; returning any 50x response that is normally returned when commands are not recognized is appropriate.

Which is exactly what `TIdFTPServer` already does. By default, it will send a `500 'ALGS': command not understood` reply.

PR #574, opened 2024-12-26
Solaris does not allow an IPv4 socket and an IPv6 socket to listen on the same port at the same time.

coffeegreg:
Solaris does not allow an IPv4 socket and an IPv6 socket to listen on the same port at the same time.
Tested with Solaris 11.4.


rlebeau (changes requested):
Indy still supports older compilers, so it can't use `#if` yet. Please rewrite the change to use separate `#ifdef` for LINUX and SOLARIS

rlebeau (approved):

PR #576, opened 2025-01-06
Adding Lazarus Package file support to Indy 10.7

MWASoftware:
This pull requests removes the out-of-date "getIndy.sh" and replaces it with a set of lazarus packages in the new subdir "lazarus-fpc". This packages have the same structure as the Delphi packages and do not require any changes to the source tree to be 
used.
Makefile.fpc files are also provided for use with fpcmake.

See README.lazarus-fpc for more information.

There are also some minor patches to the source files in order to ensure that they can be compiled with FPC.


Lib/System/IdWinsock2.pas
```diff
@@ -158,6 +158,8 @@ interface
 
 }
 
+{$IFDEF WINDOWS}
```
rlebeau, line 161: IdWinsock2 is a Windows-only unit. It should not be compiled in other platforms. This check is redundant.

Lib/System/IdWship6.pas
```diff
@@ -34,6 +34,7 @@ interface
 
 {$I IdCompilerDefines.inc}
 
+{$IFDEF WINDOWS}
```
rlebeau, line 37: IdWship6 is a Windows-only unit. It should not be compiled in other platforms. This check is redundant.

lazarus-fpc/indysystem.pas
```diff
@@ -0,0 +1,27 @@
...
+ }
+
+unit indysystem;
+
+{$warn 5023 off : no warning about unused units}
+interface
+
+uses
+  IdBaseComponent, IdComponent, IdCTypes, IdException, IdGlobal, IdIDN, 
+  IdResourceStrings, IdResourceStringsDotNet11, IdResourceStringsIconv,
```
rlebeau, line 12: FreePascal does not support Dotnet and should not be including Indy's Dotnet units.

lazarus-fpc/indysystem.pas
```diff
@@ -0,0 +1,27 @@
...
+unit indysystem;
+
+{$warn 5023 off : no warning about unused units}
+interface
+
+uses
+  IdBaseComponent, IdComponent, IdCTypes, IdException, IdGlobal, IdIDN, 
+  IdResourceStrings, IdResourceStringsDotNet11, IdResourceStringsIconv, 
+  IdResourceStringsKylixCompat, IdResourceStringsTextEncoding, 
+  IdResourceStringsUnix, IdResourceStringsVCLPosix, IdStack, IdStackBSDBase,
```
rlebeau, line 14: VCLPosix is for Delphi only. FreePascal should not be compiling VCLPosix units.

lazarus-fpc/indysystem.pas
```diff
@@ -0,0 +1,27 @@
...
+{$warn 5023 off : no warning about unused units}
+interface
+
+uses
+  IdBaseComponent, IdComponent, IdCTypes, IdException, IdGlobal, IdIDN, 
+  IdResourceStrings, IdResourceStringsDotNet11, IdResourceStringsIconv, 
+  IdResourceStringsKylixCompat, IdResourceStringsTextEncoding, 
+  IdResourceStringsUnix, IdResourceStringsVCLPosix, IdStack, IdStackBSDBase, 
+  IdStackConsts, IdStream, IdStreamVCL, IdStruct, IdTransactedFileStream, 
+  IdVCLPosixSupplemental, IdWinsock2, IdWship6, IdAntiFreezeBase,
```
rlebeau, line 16: IdWinsock2 and IdWship6 should not be compiled on non-Windows platforms.

Lib/System/IdWinsock2.pas
MWASoftware: Lazarus/fpc is a cross-platform write once environment and has a substantial number of Windows users. It is wrong to exclude a unit from a Lazarus package on the grounds that it can only be used in Windows. 

Conditional compilation within the unit is the simplest way of allowing platform specific code to be part of a unit. The Lazarus IDE maintains package units automatically and platform specific units cannot be readily included in a package whilst compiling them only when the package is compiled on their target platform.

Lib/System/IdWship6.pas
MWASoftware: Lazarus/fpc is a cross-platform write once environment and has a substantial number of Windows users. It is wrong to exclude a unit from a Lazarus package on the grounds that it can only be used in Windows. 

Conditional compilation within the unit is the simplest way of allowing platform specific code to be part of a unit. The Lazarus IDE maintains package units automatically and platform specific units cannot be readily included in a package whilst compiling them only when the package is compiled on their target platform.

lazarus-fpc/indysystem.pas
MWASoftware: I have removed the dotNet file, but I am confused by the reference to Iconv. I can't find a reference to IdIconv.pas in the lazarus packages - although there is a pointless reference to IdResourceStringsIconv,.

Lib/System/IdWinsock2.pas
JPeterMugaas: Actually, you can edit the package's .pas file and defines for platforms.

See: https://wiki.lazarus.freepascal.org/Lazarus_Packages#Platform_specific_units

MWASoftware: True - but I've had the IDE overwrite my changes to the package's .pas file before now and I would not be comfortable publishing a package that relied on conditional defines in the package's .pas file.

JPeterMugaas: I think you also have to disable "Use Unit" after selecting it in the package editor.

MWASoftware: I believe that all "Use Unit" does is to stop a using project from automatically including a dependent package's unit in its list of units. 
The underlying problem is that the Lazarus package editor does not have a means of marking a unit as valid only for certain platforms. Hence, you have to include the conditional compilation in the unit itself. On the other hand, this problem does not seem to occur with FPC packages where (manually maintained)  Makefiles are able to manage platform dependencies.

rlebeau: At least for Delphi, Indy has always relied on conditional compilation in each package's `.dpk` files, using `{$IFDEF}` blocks to control platform-specific unit references  The individual `.pas` units depend on this behavior, and thus do not `IFDEF` their entire code, as proposed in this PR.  It does not make sense to forcefully build a unit if it is not going to be used on every platform.  I'd rather not compile "empty" units.  If possible, each package should specify the specific units they actually want for each platform.

rlebeau:
> Note: I found an error in the merge changes from your upstream repo - a missing parameter in Lib/Core/IdDsnBaseCmpEdt.pas in a call to showaboutbox. I had to fix this to compile and test - hopefully added the correct param value.

`IdDsnBaseCmpEdt.pas` is a dead unit, it hasn't been touched in 15 years, and nothing in Indy refers to it.  It should be removed from the repo.  The current About dialog is shown in `IdCoreDsnRegister.pas`, which calls `TfrmAbout.ShowDlg()`, which calls `ShowAboutBox()` with the correct 3 parameter values.

PR #650, opened 2025-10-21
Missing unit after cleaning out EMB's D13 Indy and recompiling from this repo

corneliusdavid:
Indy projects compiled but working on a separate WebBroker project wouldn't, complained about missing IdHTTPWebBrokerBridge unit.

Also, ignoring .dsv files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,4 @@ __recovery/
 modules/
 
 # End of https://www.gitignore.io/api/delphi
+*.dsv


rlebeau:
`IdHTTPWebBrokerBridge.pas` is intentionally omitted from the Protocols package. It is not part of the Indy library itself, it is just an add-on unit that connects Indy to WebBroker, so only WebBroker apps would use it.  So it doesn't need to be in the Protocols package.

`IdHTTPWebBrokerBridge.pas` is shipped with the IDE, and is in this repo, so I don't know why you are missing it.

Lib/Protocols/IndyProtocols370.dpk
```diff
@@ -299,6 +299,7 @@ contains
   IdZLib in 'IdZLib.pas',
   IdZLibCompressorBase in 'IdZLibCompressorBase.pas',
   IdZLibConst in 'IdZLibConst.pas',
-  IdZLibHeaders in 'IdZLibHeaders.pas';
+  IdZLibHeaders in 'IdZLibHeaders.pas',
+  IdHTTPWebBrokerBridge in 'IdHTTPWebBrokerBridge.pas';
```
rlebeau: Omitted on purpose, not going to merge this, please remove it.

rlebeau (changes requested):
See comments

rlebeau:
In all honesty, I don't use WebBroker myself, and don't really understand why things are setup this way.  But I imagine you could simply add `IdHTTPWebBrokerBridge.pas` to your WebBroker project and compile it locally.  If you have multiple WebBroker projects then it would make sense to copy the .dcu file into your library path.

PR #653, opened 2025-11-04
add unit scope for Windows and Registry units

corneliusdavid:
TL;DR: Win64 projects don't list all needed unit scope names.

When trying to compile the Indy libraries for 64-bit, not all unit scope names were defined in the project (couldn't find the Windows or Registry units). Instead of adding unit scope names to every 64-bit package (like they are in the 32-bit packages) for all versions of Delphi (since XE2 when 64-bit compilers were introduced) or requiring Delphi developers to add them to their own Delphi options, just add compiler directives for the handful of places in the source that need it.

I found "VCL_XE2_OR_ABOVE" already defined and convenient so just used that for both Windows and Registry units.

### DIFF
--- a/Lib/Core/IdCustomTCPServer.pas
+++ b/Lib/Core/IdCustomTCPServer.pas
@@ -425,7 +425,11 @@ implementation
 uses
   {$IFDEF VCL_2010_OR_ABOVE}
     {$IFDEF WINDOWS}
-  Windows,
+      {$IFDEF VCL_XE2_OR_ABOVE}
+      WinAPI.Windows,
+      {$ELSE}
+      Windows,
+      {$ENDIF}
     {$ENDIF}
   {$ENDIF}
   IdGlobalCore,
--- a/Lib/Core/IdIOHandler.pas
+++ b/Lib/Core/IdIOHandler.pas
@@ -792,7 +792,11 @@ uses
     {$ENDIF}
   {$ENDIF}
   {$IFDEF WIN32_OR_WIN64}
-  Windows,
+    {$IFDEF VCL_XE2_OR_ABOVE}
+    WinAPI.Windows,
+    {$ELSE}
+    Windows,
+    {$ENDIF}
   {$ENDIF}
   {$IFDEF USE_VCL_POSIX}
     {$IFDEF OSX}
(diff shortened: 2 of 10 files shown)


corneliusdavid:
If you'd like some help making these changes, let me know how you want to go about it and I can put some time into it.

corneliusdavid:
On second thought, I'm going to abandon this effort. There are way too many of these; it would be far easier to add unit scope names to all the projects. The reason is that anywhere a unit name prefixes a function or procedure or identifier, it has to be a fully qualified unit name, including the unit scope name (see this [DocWiki reference](https://docwiki.embarcadero.com/RADStudio/Florence/en/Unit_Scope_Names)). That means not only do we need these `{$IFDEF USE_UNITSCOPENAMES}` conditionals in the `uses` clauses but we also need them anytime we use them in the code, such as calling `Windows.Infinite` or `SysUtils.AnsiPos` or `System.TimeSpan` (all of these and 29 more in the `IdGlobal` unit _alone_).

corneliusdavid:
Wow, you really went there:

> 315 files changed, +694 -692 lines changed

You must've used global search and replace with regular expressions or have some custom tool to make all those changes that quickly. I tried a couple of techniques but it was just taking too long so backed it all out.

I think modifying the project options would be cleaner, easier to read the code. I could fill in the missing project files--I have most versions of Delphi installed on VMs back to XE and even Delphi 7 and 2007.

However, I'll build this and see what we get. Probably better to explicitly use the unit names to remove any possibility of ambiguity or remembering to add the unit scopes in future projects.

rlebeau:
> You must've used global search and replace with regular expressions or have some custom tool to make all those changes that quickly.

Nope. I'm not good with that stuff. I just used good old fashioned Notepad++ "Find in Files" and copy & pasting.

> I think modifying the project options would be cleaner, easier to read the code.

Yup.

> I could fill in the missing project files--I have most versions of Delphi installed on VMs back to XE and even Delphi 7 and 2007.

Cool.

rlebeau:
I went ahead and merged this particular PR into the master code. We can continue with the other PR to clean things up better.

corneliusdavid:
Sounds good.

PR #654, opened 2025-11-06
More unit scope name work

corneliusdavid:
I've got a build script that compiles all the Indy libraries against several versions of Delphi and I kept iterating through them with both 32-bit and 64-bit finding several more things to fix. These submissions are the result and compile under Delphi XE, XE2, XE7, XE8, 11, 12, and 13.

### DIFF
--- a/Lib/Core/IndyCore370.dproj
+++ b/Lib/Core/IndyCore370.dproj
@@ -1,0 +1,298 @@
+﻿<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <PropertyGroup>
+        <ProjectGuid>{94A36F98-1888-4DCE-81D5-490FD57C8B99}</ProjectGuid>
+        <MainSource>IndyCore370.dpk</MainSource>
+        <Base>True</Base>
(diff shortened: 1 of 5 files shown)


Lib/Core/IdTCPClient.pas
```diff
@@ -433,6 +433,7 @@ procedure TIdTCPClientCustom.SetHost(const AValue: string);
   end;
   if GStack.IsIP(AValue) then begin
```
rlebeau: Whoops, that code wasn't supposed to be checked in. It was just something I was playing with in another branch. I'll take it out.

rlebeau:
There are a lot of files in this PR that don't actually have code changes in them, but probably just whitespace differences?

Also, I don't think the Python script should be checked in since it's not needed by the actual repo/library and it's just a one-off script.

corneliusdavid:
Sorry, it was late and I just added all files that were flagged as "changed" without looking carefully. I'm sure Claude Code made some adjustments that were not necessary; all I know is that it compiled afterwards. :-D

Yeah, the python script should've been left out.

